PR #23, opened 2020-05-01
feat: add deleting digital ocean server functionality 

Akirtovskis:
Woop woop 🎉 🎉 🎉  

First PR 

Solving : https://github.com/ledokku/ledokku/issues/8

**How to test :** 
- create server from graphiql from which will return id 
- run deleteServer mutation from graphiql with said id 
- voila your server is delete both from DigitalOcean and local db


pradel:
nice 👌

PR #25, opened 2020-05-03
feat: add basic form validation for client

Akirtovskis:
- Setup yup schema for digital ocean access token and server name
- Min and max checks were added to give more feedback if token has not pasted properly or something extra entered
- Added helper text in UI component
- Nice addition to this would be disabled buttons, perhaps todo for @bartaxyz ?

Screenshot 2020-05-03 at 10 34 38

### DIFF
--- a/client/src/pages/onboarding/cloud-provider.tsx
+++ b/client/src/pages/onboarding/cloud-provider.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useRouter } from 'next/router';
 import { useFormik } from 'formik';
-
+import * as Yup from 'yup';
 import { useSaveDigitalOceanAccessTokenMutation } from '../../generated/graphql';
 import { OnboardingLayout } from '../../layouts/OnboardingLayout';
 import withApollo from '../../lib/withApollo';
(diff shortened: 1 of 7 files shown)


client/src/pages/onboarding/cloud-provider.tsx
```diff
@@ -27,11 +27,12 @@ const CloudProvider = () => {
...
   ] = useSaveDigitalOceanAccessTokenMutation();
-  const digitalOceanAccessTokenSchema = Yup.object().shape({
-    token: Yup.string()
+  const digitalOceanAccessTokenSchema = yup.object().shape({
+    token: yup
+      .string()
       .trim()
-      .min(64, 'Token too short, it should be exactly 64 chars long')
-      .max(64, 'Token too long, it should be exactly 64 chars long')
+      .min(64, 'Token too short, it should be exactly 64 characters long')
```
pradel, line 34: @Akirtovskis the regex already say that the string should be 64 chars long, so is this really needed?

Akirtovskis: @pradel idea behind this was to give better feedback, when for example you paste incomplete token or enter something afterwards, but could be also removed. What do you think?

Akirtovskis:
@pradel server side validation added 🚀

client/src/pages/onboarding/cloud-provider.tsx
pradel: Yeah make sense 👍

pradel:
@Akirtovskis 🚀

PR #26, opened 2020-05-10
feat: setup docusaurus

Akirtovskis:
@pradel 🦖 is there! 

With initial setup there came separate `yaml` file with docusaurus service, I moved that to our `yml` and just removed the other one. It felt right and it's working, so I am assuming that was a good idea. 

Apart from that this PR seems ready to go.


pradel:
@Akirtovskis really cool!
Can we setup the v2 of the docusaurus website? https://v2.docusaurus.io/
Also as we are using a monorepo I think it make sense to create a `website` folder where docusaurus will be and add it to the workspaces array :) you can check https://github.com/accounts-js/accounts as we have the same setup

Akirtovskis:
@pradel V2 is setup and also `website` directory added to workspace array. 

V2 is absolute 🔥, looks so good.

pradel:
Sweet, next step will be to put some real data and connect it to the domain 🚀

pradel:
Do we want to use vercel or netlify to host the website?

Akirtovskis:
I have never tried Vercel, so this could be a good chance to do it. Will create todo for this.

PR #27, opened 2020-05-10
feat: run with local dokku

pradel:
Added the following graphql queries

```graphql
query {
  dokkuPlugins {
    version
    plugins {
      name
      version
    }
  }
}
```

Mutations:
```graphql
mutation {
  createApp(input: { name: "node6", gitUrl: "https://github.com/heroku/node-js-getting-started" }) {
    app {
      id
      name
    }
    appBuild {
      id
      status
    }
  }
}

```


server/src/graphql/mutations/loginWithGithub.ts
```diff
@@ -56,6 +56,13 @@ export const loginWithGithub: MutationResolvers['loginWithGithub'] = async (
 
   // User not found it means we are creating a new user
   if (!user) {
+    // We limit to only one user per server for now
+    // This can be fixed later by allowing user to invite other users for example
+    const nbUsers = await prisma.user.count();
```
Akirtovskis, line 61: @pradel just curious, what does `nb` stand for? Seems good idea to rename to something that's more explicit. What do you think?

pradel: nb stands for number :) can change it to usersNumber if it's more clear

PR #32, opened 2020-05-21
feat: add app logs query

Akirtovskis:
App logs query has arrived. 

Also moved out appName validation outside of create app fn.

### DIFF
--- a/server/src/graphql/mutations/createApp.ts
+++ b/server/src/graphql/mutations/createApp.ts
@@ -1,19 +1,9 @@
-import * as yup from 'yup';
 import { MutationResolvers } from '../../generated/graphql';
 import { prisma } from '../../prisma';
 import { dokku } from '../../lib/dokku';
 import { buildAppQueue } from '../../queues/buildApp';
 import { sshConnect } from '../../lib/ssh';
-
-// Validate the name to make sure there are no security risks by adding it to the ssh exec command.
-// Only letters and "-" allowed
-// TODO unit test this schema
-const createAppSchema = yup.object({
-  name: yup
-    .string()
-    .required()
-    .matches(/^[a-z0-9-]+$/),
-});
+import { appNameSchema } from '../utils';
 
 export const createApp: MutationResolvers['createApp'] = async (
   _,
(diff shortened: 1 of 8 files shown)


server/src/graphql/mutations/createApp.ts
```diff
@@ -1,19 +1,9 @@
...
 import { MutationResolvers } from '../../generated/graphql';
 import { prisma } from '../../prisma';
 import { dokku } from '../../lib/dokku';
 import { buildAppQueue } from '../../queues/buildApp';
 import { sshConnect } from '../../lib/ssh';
-
-// Validate the name to make sure there are no security risks by adding it to the ssh exec command.
-// Only letters and "-" allowed
-// TODO unit test this schema
-const createAppSchema = yup.object({
```
pradel, line 11: @Akirtovskis the schema will really soon be different, it's a good idea to share the name validation between the mutations but maybe we can do something like:
```
const createAppSchema = yup.object({
 name: importedFromFile,
});
```
What do you think?

server/src/lib/dokku/apps/logs.ts
```diff
@@ -0,0 +1,20 @@
...
+  const allLogs = commandResult.split('\u001b');
+  //We remove redundnat chars from good log & pick only 3rd item in array
+  //which is a good log
+  const logs = allLogs[2].slice(4);
+  return logs;
+};
+
+export const logs = async (ssh: NodeSsh, name: string) => {
+  const resultAppLogs = await ssh.execCommand(`logs ${name}`);
+  console.log(resultAppLogs);
```
pradel, line 13: can you remove this one? :)

server/src/graphql/mutations/createApp.ts
Akirtovskis: @pradel what does `importedFromFile` mean in this case? Is it a separate yup schema that we will be exporting from somewhere?

pradel: Yup eg:
```
# utils.ts
const importedFromFile =  yup
    .string()
    .required()
    .matches(/^[a-z0-9-]+$/)

# createApp.ts
const createAppSchema = yup.object({
 name: importedFromFile,
});
```

Akirtovskis: Sounds great, will do.

Akirtovskis: Done.

PR #34, opened 2020-05-22
feat: add app deletion mutation

Akirtovskis:
This works, but couple of comments/questions : 

1. Currently using `--force destroy` which seems fine as we can handle more validations on client side, for example popup "to confirm enter name of the app". 

2. Do we need queue for app deletion? If so then it will be a fun tech task to build one :) 

3. I didn't parse `stdout` because it didn't seem to provide any valuable log that we wouldn't be able to make ourselves. Here is now it looked :  
`stdout: '-----> Destroying node14 (including all add-ons)\n-----> Cleaning up...',"`


server/src/graphql/mutations/deleteApp.ts
```diff
@@ -0,0 +1,63 @@
...
+  );
+
+  await prisma.appBuild.deleteMany({
+    where: {
+      id: appBuildToDelete[0].id,
+    },
+  });
+
+  // We delete the app itself
+  await prisma.app.delete({
```
pradel, line 56: @Akirtovskis do you know if prisma support cascade delete? If yes that would simplify a lot this function

pradel:
1. Yup good idea :+1: 
2. Well depends how long it is to delete an app?
3. Will depends of the response to the énd question :D

server/src/graphql/mutations/deleteApp.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  // We make sure the name is valid to avoid security risks
+  appNameSchema.validateSync({ name: input.name });
+
+  // TODO check name of the app is unique per server
+
+  const ssh = await sshConnect();
+
+  await dokku.apps.deleteApp(ssh, input.name);
```
pradel, line 23: we probably want to do the find one on the db before calling dokku :)

server/src/graphql/mutations/deleteApp.ts
Akirtovskis: @pradel they currently don't.  It seems that it will be done at some point "soon" which probably means this year as discussed here : https://github.com/prisma/migrate/issues/249 . 

There's a [link](https://prisma2.netlify.app/guides/database-workflows/cascading-deletes/postgresql/#3-create-two-tables-with-a-foreign-key-and-cascade-deletion-behavior) to describe how it can be done, but for me it seems better to wait while they implement this feature. So I added to do for now. 

What's your opinion on this? Better to wait or go with what was suggested in the link?

pradel: Let's wait and change the logic here once it's ready :+1:

PR #36, opened 2020-05-23
feat: add env variables

Akirtovskis:
Whoop whoop 🎉 , this works! 

Couple of comments and questions : 

1. Editing works as you suggested, add(ENVVAR_KEY=NEW_VALUE) will overwrite (ENVVAR_KEY=OLD_VALUE), so no extra mutation needed. 

2. Env var change seems to be redeploying app, but as far as I checked, that doesn't seem to kill it, so while the result is not there yet, we could perhaps show something on UI or mention before, that it takes 1 min for new env_vars to take effect. What do you think? 

3. `addEnvVars` and `deleteEnvVars` mutations currently return string with short info about affected env_vars. Is this good or should we move it to boolean and then on UI handle info based on that boolean? 

4. Future todos that came to my mind : most common dokku env var documentation, advances feature where you can just see input window and set multiple env vars with just one click. I.e pasting 1 key value pair in each line, so it's possible to parse it.


server/src/index.ts
```diff
@@ -76,6 +93,17 @@ const typeDefs = gql`
     type: DatabaseTypes!
   }
 
+  input AddEnvVarInput {
+    name: String!
```
pradel, line 97: For both delete and add on the ui we will have acces to the appId only

server/src/index.ts
```diff
@@ -86,12 +114,15 @@ const typeDefs = gql`
     databases: [Database!]!
     dokkuPlugins: DokkuPluginResult!
     appLogs(name: String!): AppLogsResult!
+    envVars(name: String!): EnvVarsResult!
```
pradel, line 117: I just saw that but actually we will have access to appId instead of app name, will be the same `appLogs`.

server/src/graphql/queries/envVars.ts
```diff
@@ -0,0 +1,22 @@
...
+  { name },
+  { userId }
+) => {
+  if (!userId) {
+    throw new Error('Unauthorized');
+  }
+
+  appNameSchema.validateSync({ name });
+
+  const ssh = await sshConnect();
```
pradel, line 17: before this part for the query and both mutations we should do a findOne to check that the app exist in the db first, if does not exist, throw an error

server/src/lib/dokku/index.ts
```diff
@@ -2,9 +2,13 @@ import { create } from './apps/create';
...
 import { logs } from './apps/logs';
+import { listVars } from './env/listVars';
+import { add } from './env/add';
 import { deleteApp } from './apps/delete';
+import { deleteEnvVar } from './env/delete';
 
 export const dokku = {
   apps: { create, logs, deleteApp },
   plugin: { installed, list },
+  env: { listVars, add, deleteEnvVar },
```
pradel, line 13: I think we should try to map the structure to the dokku commands, in that case instead of `env` call it config

server/src/index.ts
Akirtovskis: So flow is like this : 
- we pass appId, 
- we get app name in our db from the appId, 
- we pass it to dokku cli. 

Is this correct?

server/src/lib/dokku/index.ts
Akirtovskis: This makes sense, should we do also for `set` and `unset` vars as well?

pradel:
Editing works as you suggested, add(ENVVAR_KEY=NEW_VALUE) will overwrite (ENVVAR_KEY=OLD_VALUE), so no extra mutation needed.

Env var change seems to be redeploying app, but as far as I checked, that doesn't seem to kill it, so while the result is not there yet, we could perhaps show something on UI or mention before, that it takes 1 min for new env_vars to take effect. What do you think?

addEnvVars and deleteEnvVars mutations currently return string with short info about affected env_vars. Is this good or should we move it to boolean and then on UI handle info based on that boolean?

Future todos that came to my mind : most common dokku env var documentation, advances feature where you can just see input window and set multiple env vars with just one click. I.e pasting 1 key value pair in each line, so it's possible to parse it.

1. That's really great
2. I think this should be added as a background task no? Currently calling the mutation should trigger a new task that will redeploy the app automatically (done automatically by dokku but thuis can take some time for big apps so we can't block the http request for that long)
3. I guess boolean make more sense here yeah :)
4. Like a `.env` file?

server/src/index.ts
pradel: Yup :+1:

server/src/lib/dokku/index.ts
pradel: would be great :)

server/src/graphql/mutations/deleteApp.ts
```diff
@@ -2,7 +2,6 @@ import { MutationResolvers } from '../../generated/graphql';
 import { prisma } from '../../prisma';
 import { dokku } from '../../lib/dokku';
```
Akirtovskis, line 3: In this file messed up something, currently fixing 🏇

Akirtovskis:
@pradel all the previous feedbacks fixed.

Question about 2. , how do I create a background task? 

4. Yes, like pasting .env file instead of manually entering it field by field.

pradel:
@Akirtovskis to create a background task we use bull.

For inspiration you can take a look at how we do to create the app.
- Background task to create an app server/src/queues/buildApp.ts
- The task is triggered here server/src/graphql/mutations/createApp.ts

I guess for the first version we can do it quite simple (no need for sockets.io), basically the task will just run the ssh command.

server/src/queues/setEnvVar.ts
```diff
@@ -0,0 +1,48 @@
+import { Worker, Queue } from 'bullmq';
```
Akirtovskis, line 1: @pradel my first attempt at creating queue! It works as I understood it should :  mutation now takes almost no time and env vars are set correctly when I check them. And currently we don't produce any logs or other feedback, but as I understood that was the plan for this one. 

If this is good, I will create one for `unsetEnvVar` as well.

pradel: @Akirtovskis Yup you are right, I guess the best will be to notify the user that the app will be redeployed when the env is changing and add a new entry in the activity list :)

pradel: Looks good!

Akirtovskis: @pradel unsetEnvVar queue added, all the feedbacks are fixed and naming is improved to map to 🐳 . Should be good to go, but I am happy to fix if something is still no bueno.

server/src/lib/dokku/apps/delete.ts
```diff
@@ -1,13 +1,13 @@
 import NodeSsh from 'node-ssh';
 
-// named deleteApp instead of delete because of the "strict" warning
-// apparently you can't use word "delete" as var name
-export const deleteApp = async (ssh: NodeSsh, appName: string) => {
-  const resultAppsDelete = await ssh.execCommand(
+export const destroy = async (ssh: NodeSsh, appName: string) => {
```
pradel, line 3: should the name of the file change to destroy too? (not a blocker for merging can be done later :) )

PR #38, opened 2020-05-27
 feat(client): app view

Akirtovskis:
Sneak peak on how individual app view could look. 

I think there's lots to iterate on, but posting it here so we have something to begin with. 

**App view**

Screenshot 2020-05-27 at 23 34 03

**Env section**

Screenshot 2020-05-29 at 08 04 26


Akirtovskis:
@pradel env_var form logic instinctively seems messy to me, so if there are any better ways this could be done, I am open to improving it, other than that this is ready to go.

pradel:
@Akirtovskis it looks good, if it's working let's go with this for now, we can still improve it later :)

PR #44, opened 2020-05-29
feat(client): add app settings view

Akirtovskis:
@pradel Whoop whoop! Another view bites the dust 🎉 . 

Works as expected, one thing that could probably be fixed is adding refetchQuery when redirected to dashboard. Currently when we first get to dashboard for a short bit deleted app still appears there. 

W/o apollo hooks and `useMutation(MUTATION_NAME)` you just do refetchQueries, but how do I handle this when it's `useThisParticularMutation()`? 

Screenshot 2020-05-29 at 21 58 07

### DIFF
--- a/client/src/generated/graphql.tsx
+++ b/client/src/generated/graphql.tsx
@@ -229,6 +229,19 @@ export type CreateDatabaseMutation = (
   ) }
 );
 
+export type DestroyAppMutationVariables = {
+  input: DestroyAppInput;
+};
+
+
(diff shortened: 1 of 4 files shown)


pradel:
You can pass refetchQueries directly to the options iirc

Akirtovskis:
Ok, will try to make this work tomorrow. 💤

Akirtovskis:
@pradel fixed and good to 🏁 .

PR #47, opened 2020-05-30
feat(client): fix newly created app issues

Akirtovskis:
Before this branch, we should merge : https://github.com/ledokku/ledokku/pull/46 . 

Multiple issues fixed : 
- refetch dashboard query after new app creation 
- when immediately click on new app, app view works with the "app is still deploying log", which 
changes to 🟢 on the first poll after 2.5 min. This was super exciting to see for the first time. 👀  

Screenshot 2020-05-30 at 08 03 57

### DIFF
--- a/client/src/pages/app/[appId]/index.tsx
+++ b/client/src/pages/app/[appId]/index.tsx
@@ -30,7 +30,7 @@ const App = () => {
     variables: {
       appId,
     },
-    ssr: true,
+    ssr: false,
     skip: !appId,
     // we fetch status every 2 min 30 sec
     pollInterval: 15000,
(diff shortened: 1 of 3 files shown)


pradel:
@Akirtovskis super cool stuff let's resolve the conflicts and merge!

client/src/pages/app/[appId]/index.tsx
```diff
@@ -28,10 +28,10 @@ const App = () => {
     variables: {
       appId,
     },
-    ssr: false,
+    ssr: true,
```
Comment (pradel) [inline client/src/pages/app/[appId]/index.tsx:31]: for now this should stay false :)

PR #49, opened 2020-05-30
chore(server): migrate createdAt fields from datetime to string

Akirtovskis:
Now we see real app creation date instead of hardcoded one.

@pradel currently I created migration only for app and appBuild. 

I guess couple of questions regarding this : 
- What would be the benefits of leaving DateTime in prisma schema vs moving it to String? 
- Should I convert all models to String for now or the plan is to sometime soon create custom graphql scalar DateTime?

This is the last PR today 🤣

### DIFF
--- a/client/src/generated/graphql.tsx
+++ b/client/src/generated/graphql.tsx
@@ -20,6 +20,7 @@ export type App = {
   id: Scalars['ID'];
   name: Scalars['String'];
   githubRepoUrl: Scalars['String'];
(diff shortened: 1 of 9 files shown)


pradel:
> What would be the benefits of leaving DateTime in prisma schema vs moving it to String

Well if you move them to string you will lose the ability to do time based queries, the prisma schema should stay unchanged and we have to keep the DateTime type :)

PR #53, opened 2020-05-31
chore(server): add graphql custom scalars

Akirtovskis:


client/src/generated/graphql.tsx
```diff
@@ -11,15 +11,19 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
+  /** Use JavaScript Date object for date/time fields. */
+  DateTime: any;
```
pradel, line 15: instead of any we could type it by adding the following property in the codegen file
```
    config:
      scalars:
        DateTime: string
```

server/src/generated/graphql.ts
```diff
@@ -10,15 +10,19 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
+  /** Use JavaScript Date object for date/time fields. */
+  DateTime: any;
```
pradel, line 14: instead of any we could type it by adding the following property in the codegen file
```
    config:
      scalars:
        DateTime: Date
```

Akirtovskis:
@pradel good to 🚀

client/codegen.yml
```diff
@@ -13,3 +13,5 @@ generates:
       withHOC: false
       withHooks: true
       enumsAsTypes: true
+      scalars:
+        DateTime: String
```
pradel, line 17: @Akirtovskis it should actually be `string` lowercase to represent the typescript type :)

pradel:
Once the type is fixed, good to go 🚀

PR #63, opened 2020-06-05
feat(client): move specific app db settings out of app info view

Akirtovskis:


pradel:
@Akirtovskis is this one good to go?

Akirtovskis:
@pradel Yes, this is good to go.

PR #70, opened 2020-06-07
feat(client): delete-env-var

Akirtovskis:
Screenshot 2020-06-07 at 21 21 19


client/src/pages/app/[appId]/env.tsx
```diff
@@ -54,56 +69,71 @@ const EnvForm = ({ name, value, appId, isNewVar }: EnvFormProps) => {
...
+              <button
+                type="submit"
+                className="inline py-2 px-10 bg-gray-900 hover:bg-blue text-white  font-bold hover:text-white border hover:border-transparent rounded-lg"
+              >
+                {isNewVar ? 'Add' : 'Save'}
+              </button>
+            </div>
+          </div>
+        </form>
+        {!isNewVar && (
```
Comment (pradel) [inline client/src/pages/app/[appId]/env.tsx:125]: I think it would make more sense to add the buttons together inside the form no?

Comment (pradel) [reply client/src/pages/app/[appId]/env.tsx:125]: inside `handleDeleteEnvVar` you can user `event.preventDefault()` to cancel the form submission

Comment (Akirtovskis) [reply client/src/pages/app/[appId]/env.tsx]: Yes, makes sense, fixed 🚀 .

PR #71, opened 2020-06-13
feat(client): create button component

Akirtovskis:
Only issue I was facing with the component I created was whenever button needs to be used within Link component, it should be wrapped with <a> tag to work properly. Apart from that everything else seems to work.


client/src/ui/components/Button.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import cx from 'classnames';
+
+interface ButtonProps {
```
pradel, line 4: the best would be to extend the button props and forward them to the button using `{...props}` so we don't have to redefine onClick etc..

client/src/ui/components/Button.tsx
```diff
@@ -0,0 +1,50 @@
...
+    iconStart,
+    color,
+    disabled,
+    className,
+    size,
+    isSubmit,
+    onClick,
+  } = props;
+  return (
+    <button
```
pradel, line 28: we should pass the disabled prop too otherwise if disabled is true, you can still click on the button

client/src/ui/components/Button.tsx
Akirtovskis: @pradel  resolved both feedbacks in this one.

pradel: @Akirtovskis on this one I meant `interface ButtonProps extends React.HTMLProps<HTMLButtonElement>`, so that way we don't have to define onClick at all :)

Akirtovskis: @pradel is this resolved or still in question? It didn't work out of the box like you suggested, so I assume that's why it is resolved.

client/src/ui/components/Button.tsx
```diff
@@ -0,0 +1,44 @@
...
+  disabled,
+  width,
+  className,
+  iconEnd,
+  iconStart,
+  ...props
+}: ButtonProps) => (
+  <button
+    {...props}
+    type={isSubmit ? 'submit' : 'button'}
```
pradel, line 25: It would be cleaner to just use the standard `type` prop and remove the isSubmit prop :)

PR #73, opened 2020-06-14
feat: setup docker release process

pradel:
The ci/cd is now building the docker images
- latest is the master branch
- create a new tag on every new release
- expose env variables to client at runtime
- run prisma migrations when server start on prod


.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
+name: Docker
+
+on:
```
pradel, line 3: remove this before merging

website/docs/installation.md
```diff
@@ -0,0 +1,89 @@
...
+# If the Postgres plugin is not installed on your server run, otherwise skip this command
+sudo dokku plugin:install https://github.com/dokku/dokku-postgres.git postgres
+
+dokku postgres:create ledokku
+dokku postgres:link ledokku ledokku
+```
+
+In your browser open https://github.com/settings/developers and click on the "New OAuth App" button.
+
+Add a name, a homepage url, and in the field "Authorization callback URL" set the value to your ledokku ip address and port 4000. (eg: http://123.123.123.123:4000)
```
Akirtovskis, line 39: @pradel maybe here we can add example of homepage url as well? As for someone setting everything up for the first time it could not be obvious.

PR #74, opened 2020-06-16
fix(client): add publicRuntimeConfig to nextjs config

Akirtovskis:
@pradel as far as I understood, this will solve the issue. 

Only confusing thing to me was that in docs this was mentioned : 
`Runtime configuration won't be available to any page (or component in a page) without **getInitialProps.**` . 

I tried logging the vars, logging in/out and all seemed to work. Am I missing something?


pradel:
@Akirtovskis could it be that it's the case only when doing ssr? I never used this feature before so not sure how it is working 🤔

PR #79, opened 2020-06-18
feat(client): create spinner & add to login logic

Akirtovskis:


client/src/ui/components/Spinner.tsx
```diff
@@ -0,0 +1,38 @@
...
+import cx from 'classnames';
+
+interface SpinnerProps {
+  className?: string;
+  size: 'small' | 'large';
+}
+
+export const Spinner = ({ className, size }: SpinnerProps) => (
+  <React.Fragment>
+    <style>{`.loader {
```
pradel, line 11: Using style like this is not the best way, we should rather use styled components in that case to define the animations :)

client/src/pages/home.tsx
```diff
@@ -63,7 +66,9 @@ export const Home = () => {
       {/* TODO display spinner if query is loading */}
       {loading && <p className="mt-3">Loading...</p>}
```
Akirtovskis, line 67: @Akirtovskis add spinner here as well

Akirtovskis:
@pradel this seems good to go as well.

client/src/ui/components/Spinner.tsx
```diff
@@ -1,38 +1,33 @@
...
-        {
-          'border-8 border-t-8 h-24 w-24': size === 'large',
-          'border-4 border-t-4 h-8 w-8': size === 'small',
-        },
-        className
-      )}
-    ></div>
-  </React.Fragment>
-);
+export const Spinner = (props: SpinnerProps) => <StyledSpinner {...props} />;
```
pradel: @Akirtovskis to mix both we could do something like this:
```
export const Spinner = (props: SpinnerProps) => <StyledSpinner className={cx(
        'loader ease-linear rounded-full border-gray-200 ',
        {
          'border-8 border-t-8 h-24 w-24': size === 'large',
          'border-4 border-t-4 h-8 w-8': size === 'small',
        },
        className
      )} {...props} />;
```

What do you think?

PR #80, opened 2020-06-22
feat(client): isPluginInstalled query and UI

Akirtovskis:
@pradel this is the PR in question.  Both server & client queries work and as you can see it is reusing the `plugin:list` command logic because `installed:plugin name` command also requires sudo to be run. 

As for UI, many battles were fought, but this solution works when switching between dbs. One thing that is left to figure out is how to handle when they copy and run the command. Db input field needs to be somehow displayed instead of terminal. 

Would be nice if you checked the current UI part and gave me feedback and on functionality and looks. Then I can either proceed further or improve current logic.


server/src/graphql/queries/isPluginInstalled.ts
```diff
@@ -0,0 +1,23 @@
...
+  { pluginName },
+  { userId }
+) => {
+  if (!userId) {
+    throw new Error('Unauthorized');
+  }
+
+  const ssh = await sshConnect();
+
+  const dokkuPlugins = await dokku.plugin.list(ssh);
```
pradel, line 16: you can use server/src/lib/dokku/plugin/installed.ts that will already return a boolean, will be faster ;)

pradel:
@Akirtovskis looks good and works :ok_hand: 

Just one thing, dont know if we want to do it in this pr or a following one but the mutation to create a database is actually calling the enabled command (which is obviously failing now). So currently it's not possible to create a database :D we will have to change it to your list implementation.

Akirtovskis:
@pradel what do you mean by `enabled command`? 

If the fact that db creation currently does not work is the only blocker for this then I suggest that we merge and I will make sure db creation works in the first db setup PR.

pradel:
@Akirtovskis I mean the pugin:enabled command (failing because you must be root to run it)

Akirtovskis:
@pradel ah I got it now. You probably mean `installed`.  Let's merge this one and I will solve the issue in the first db pr.

PR #88, opened 2020-07-14
feat(client,server): basic database management views

Akirtovskis:
To test - create db and navigate to different views in db management section.


server/src/graphql/mutations/createDatabase.ts
```diff
@@ -24,29 +25,25 @@ export const createDatabase: MutationResolvers['createDatabase'] = async (
   }
 
   // TODO implement other db support
-  if (input.type !== 'POSTGRESQL') {
```
pradel, line 27: we still need this check until we support more db right?

client/src/ui/components/Spinner.tsx
```diff
@@ -30,6 +30,7 @@ export const Spinner = ({ size, className, ...props }: SpinnerProps) => (
       {
         'border-8 border-t-8 h-24 w-24': size === 'large',
         'border-4 border-t-4 h-8 w-8': size === 'small',
+        'mt-1 border-2 border-t-2 h-4 w-4': size === 'extraSmall',
```
pradel, line 33: mt-1 should probably be removed from here, the other sizes don't have any margin

client/src/pages/database/env.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  return (
+    <div>
+      <Header />
+      <div className="max-w-5xl mx-auto px-4 sm:px-6 lg:px-8">
+        <TabNav>
+          <TabNavLink to={`/database/${database.id}`}>Database</TabNavLink>
+          <TabNavLink to={`/database/${database.id}/apps`}>Apps</TabNavLink>
+          <TabNavLink to={`/database/${database.id}/env`} selected>
```
pradel, line 43: do we need this one? We can't set any env for the database no?

Akirtovskis:
@pradel thanks for feedbacks, all fixed! 🚀

pradel:
Nice good job!

PR #90, opened 2020-07-14
feat(client,server): database deletion functionality

Akirtovskis:
Can be merged after : https://github.com/ledokku/ledokku/pull/88 . 

To test try creating db and deleting in the database views settings part.


pradel:
@Akirtovskis conflicts on this one :grin:

Akirtovskis:
@pradel and they are gone! :)

server/src/index.ts
```diff
@@ -208,3 +218,10 @@ http.listen({ port: 4000 }, () =>
     `🚀 Server ready at http://localhost:4000${apolloServer.graphqlPath}`
   )
 );
+
+const test = async () => {
```
pradel, line 222: let's remove this?

server/src/graphql/mutations/destroyDatabase.ts
```diff
@@ -0,0 +1,42 @@
...
+  { input },
+  { userId }
+) => {
+  if (!userId) {
+    throw new Error('Unauthorized');
+  }
+
+  const { databaseId } = input;
+
+  // TODO check name of the database is unique per server
```
pradel, line 17: This TODO can be removed

server/src/graphql/mutations/destroyDatabase.ts
```diff
@@ -0,0 +1,42 @@
...
+  const ssh = await sshConnect();
+
+  // We delete the database
+  await prisma.database.delete({
+    where: {
+      id: databaseId,
+    },
+  });
+
+  const result = await dokku.postgres.destroy(ssh, databaseToDelete.name);
```
pradel, line 39: I guess it make more sense to move this one before the prisma call. That way if dokku fail to delete the db, it's not removed from the database :)

server/src/index.ts
Akirtovskis: 🤦 for leaving this in active PR.

PR #101, opened 2020-07-19
feat(client): restructure app and db management views

Akirtovskis:
Logs moved out to separate view and connect app/db now is in the main view.

PR #103, opened 2020-07-19
feat(client, server): add db logs and refactor db server side

Akirtovskis:
Can be only merged after : https://github.com/ledokku/ledokku/pull/101 
as logs views from there are used.

Will resolve : 
https://github.com/ledokku/ledokku/issues/22
https://github.com/ledokku/ledokku/issues/21
https://github.com/ledokku/ledokku/issues/20

@pradel this pr adds logs for db view and also I realised that we can easily tweak db server side code to make all dbs work right away. 

It works because for most common db plugins commands are exactly the same. 

What do you think about this approach?


Akirtovskis:
Screenshot 2020-07-20 at 18 25 36

**AND**

Screenshot 2020-07-20 at 18 33 00

Akirtovskis:
@pradel a bit of a scroll, but for now will do the job, I think.

localhost_3000_database_d35141c3-5b82-4bb3-8224-833180e20f5f_logs (1)

server/src/graphql/queries/databaseInfo.ts
```diff
@@ -0,0 +1,33 @@
...
+    throw new Error('Unauthorized');
+  }
+
+  const database = await prisma.database.findOne({
+    where: {
+      id: databaseId,
+    },
+  });
+
+  if (!database) {
```
pradel, line 22: c

server/src/graphql/queries/databaseLogs.ts
```diff
@@ -0,0 +1,33 @@
...
+    throw new Error('Unauthorized');
+  }
+
+  const database = await prisma.database.findOne({
+    where: {
+      id: databaseId,
+    },
+  });
+
+  if (!database) {
```
pradel, line 22: We need to check that database belongs to the user

Akirtovskis:
@pradel both feedbacks fixed!

PR #110, opened 2020-08-04
feat: display linked apps/dbs

Akirtovskis:
@pradel this is also good to go. 

I think it's better to do unlink task in separate PR as it will probably be quite a big one. 

With the file changes from here, we can now see linked apps/dbs and in dropdown we only see those which are not connected to app or db. 

Couple of screenshots below : 

image

image


yarn.lock
```diff
@@ -4331,6 +4331,28 @@ apollo-server-express@2.16.1, apollo-server-express@^2.16.1:
     subscriptions-transport-ws "^0.9.16"
     type-is "^1.6.16"
 
+apollo-server-express@^2.15.1:
```
pradel, line 4334: this is pretty weird you probably want to revert :)

server/src/index.ts
```diff
@@ -166,6 +174,8 @@ const typeDefs = gql`
     appLogs(appId: String!): AppLogsResult!
     databaseInfo(databaseId: String!): DatabaseInfoResult!
     databaseLogs(databaseId: String!): DatabaseLogsResult!
+    databasesLinkedToApp(appId: String!): DatabasesLinkedToAppResult!
```
pradel, line 177: I am wondering if it should rather be a subset od apps eg:
```graphql
app(id: "") {
  id
  name
  databases {
	id
  }
}
```
what do you think?

pradel: that way we can do it in just 1 query

client/src/pages/database/index.tsx
```diff
@@ -152,6 +176,9 @@ export const Database = () => {
                   placeholder={selectedApp}
                   isSearchable={false}
                   aria-labelledby="app-select-dropdown"
+                  noOptionsMessage={() =>
```
pradel, line 179: should we just remove the select in such case? Not sure which one is the best

server/src/index.ts
Akirtovskis: @pradel this makes a lot of sense, same also could be done for the db query. With this we will be able to get rid of two extra queries. I Will refactor in this PR.

client/src/pages/database/index.tsx
Akirtovskis: I will try both approaches and see which looks better. I guess it depends on how it "feels" when you unlink db and see the select appear again. If that looks ok, then removing select and showing some text in case of no items makes sense.

Akirtovskis:
@pradel this is good to go now, with all the refactored queries. :)

server/src/graphql/queries/database.ts
```diff
@@ -10,6 +10,9 @@ export const database: QueryResolvers['database'] = async (
     throw new Error('Unauthorized');
   }
 
-  const database = await prisma.database.findOne({ where: { id: databaseId } });
+  const database = await prisma.database.findOne({
```
pradel, line 13: I think this should rather be in a resolver. That way it's not executed if you don't ask for the subfields :)

server/src/graphql/queries/app.ts
```diff
@@ -6,6 +6,11 @@ export const app: QueryResolvers['app'] = async (_, { appId }, { userId }) => {
     throw new Error('Unauthorized');
   }
 
-  const app = await prisma.app.findOne({ where: { id: appId } });
+  const app = await prisma.app.findOne({
```
pradel, line 9: I think this should rather be in a resolver. That way it's not executed if you don't ask for the subfields :)

server/src/index.ts
```diff
@@ -195,6 +198,7 @@ const apolloServer = new ApolloServer({
   typeDefs,
   resolvers: {
     ...resolvers,
+    ...customResolvers,
```
pradel, line 201: to have better type checking let's add it after line 194

server/src/graphql/resolvers.ts
```diff
@@ -0,0 +1,36 @@
...
+    },
+  },
+  Database: {
+    apps: async (database: Database) => {
+      const apps = await prisma.database
+        .findOne({
+          where: {
+            id: database.id,
+          },
+          select: {
```
pradel, line 28: do you need the select as you already use `.apps()`?

Akirtovskis:
@pradel fixed both feedbacks. 

I didn't know that `select` === `.apps()`, but now I know :)

Issue #109, opened 2020-08-04
App logs for Ruby app


Akirtovskis:
After deploying ruby heroku starter I realised that ruby logs are different from node js logs. 

So we should create different log parser  for Ruby: 
`server/src/lib/dokku/apps/logs.ts` 

For now it's just Ruby, but we need to check other languages too and potentially create issues.

pradel:
@Akirtovskis do you have an example of some ruby logs? I guess the logs output depends of the logger you utilise not the language, so we will have to find a generic way to parse the logs probably...

Akirtovskis:
@pradel there you go. This is the same command and outputs vary by quite a bit :) 

RUBY

<img width="812" alt="Screenshot 2020-08-04 at 23 11 37" src="https://user-images.githubusercontent.com/25903618/89340114-e64e6000-d6a7-11ea-9d5a-44750387ab20.png">

NODE 

<img width="500" alt="Screenshot 2020-08-04 at 23 12 21" src="https://user-images.githubusercontent.com/25903618/89340179-01b96b00-d6a8-11ea-8f66-6085625237c6.png">

Akirtovskis:
GO logs preview 

![image](https://user-images.githubusercontent.com/25903618/89655532-ee3e1800-d8d2-11ea-8bb7-ed4da40b5692.png)

Akirtovskis:
@Akirtovskis parse logs - one new log per line.

PR #113, opened 2020-08-16
fix: app logs display issue

Akirtovskis:
This PR will close : https://github.com/ledokku/ledokku/issues/109 

Added reasonably well formatted logs for all app types.

I will create issue for separating `date` and `log` that way we can display it even more beautifully and do custom colour for date and the log itself.

image


server/src/index.ts
```diff
@@ -89,7 +89,7 @@ const typeDefs = gql`
   }
 
   type AppLogsResult {
-    logs: String!
+    logs: [String]!
```
pradel, line 92: for better type checking
```suggestion
    logs: [String!]!
```

server/src/lib/dokku/apps/logs.ts
```diff
@@ -1,10 +1,18 @@
 import NodeSsh from 'node-ssh';
 
 const parseLogsCommand = (commandResult: string) => {
-  const allLogs = commandResult.split('\u001b');
+  const allLogs = commandResult.split('\n');
+  const logs = [];
   //We remove redundnat chars from good log & pick only 3rd item in array
   //which is a good log
```
pradel, line 7: should we change the comment?
Can you explain the changes you did? Like this is pretty hard to understand :D

server/src/lib/dokku/apps/logs.ts
```diff
@@ -1,10 +1,20 @@
...
 
 const parseLogsCommand = (commandResult: string) => {
-  const allLogs = commandResult.split('\u001b');
-  //We remove redundnat chars from good log & pick only 3rd item in array
-  //which is a good log
-  const logs = allLogs[2].slice(4);
+  // We split logs into array by new line
+  const allLogs = commandResult.split('\n');
+  const logs = [];
+  allLogs.map((log) => {
```
pradel, line 7: as you are using .map either you have to do `logs = allLogs.map((log) => {` or use a forEach loop

PR #115, opened 2020-08-16
fix: add isLoading prop to Button component

Akirtovskis:


client/src/ui/components/Button.tsx
```diff
@@ -44,7 +47,15 @@ export const Button = ({
     )}
   >
     {iconStart && <span className="mr-3">{iconStart}</span>}
-    {children}
+    {isLoading ? (
+      <React.Fragment>
+        <Spinner size="extraSmall" />
+        <span className="ml-2 opacity-50"> {children}</span>
```
pradel, line 53: there is a space in the `span` I think it's better to remove it

client/src/ui/components/Button.tsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import cx from 'classnames';
+import { Spinner } from './Spinner';
 
 interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
   iconStart?: React.ReactNode;
   iconEnd?: React.ReactNode;
   color: 'red' | 'grey';
   variant?: 'solid' | 'outline';
+  isLoading?: boolean;
```
pradel, line 10: Wondering if it makes sense to set `disabled` to true automatically if `isLoading` is true

client/tailwind.config.js
```diff
@@ -8,6 +8,11 @@ module.exports = {
         sans: ['Inter var', ...defaultTheme.fontFamily.sans],
       },
     },
+    minWidth: {
```
pradel, line 11: this can be reverted as we don't use it anymore :)

client/src/ui/components/Button.tsx
Akirtovskis: @pradel this is indeed a very good idea. Updated and refactored where it was necessary.

pradel (approved):

PR #120, opened 2020-09-12
feat: display info message if app is not deployed

pradel:


pradel:
We can now continue with this one https://github.com/dokku/dokku/issues/4134#issuecomment-731845451.
The best would be:
- when we sync, check if the app is deployed
- save it in the db as a boolean flag
- on the page along with the message add a button to update the app status

josegonzalez:
This should be unblocked. Also, exit codes are now slightly different (or should be) if an app does not exist. See https://dokku.github.io/release/dokku-0.24.0 for details, but basically if you get an exit code of 20, that means the app does not exist. Might save on ssh commands :)

PR #129, opened 2020-09-24
chore(client): update from apollo-client to @apollo/client

Akirtovskis:
Upgraded to @apollo/client mostly based on this: 
https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/

### DIFF
--- a/client/src/generated/graphql.tsx
+++ b/client/src/generated/graphql.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable */
 import gql from 'graphql-tag';
-import * as ApolloReactCommon from '@apollo/react-common';
-import * as ApolloReactHooks from '@apollo/react-hooks';
+import * as ApolloReactCommon from '@apollo/client';
+import * as ApolloReactHooks from '@apollo/client';
 export type Maybe<T> = T | null;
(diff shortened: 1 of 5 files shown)


client/src/generated/graphql.tsx
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable */
 import gql from 'graphql-tag';
-import * as ApolloReactCommon from '@apollo/react-common';
-import * as ApolloReactHooks from '@apollo/react-hooks';
+import * as ApolloReactCommon from '@apollo/client';
+import * as ApolloReactHooks from '@apollo/client';
```
pradel, line 4: you can replace the 2 import with `import { ApolloProvider } from '@apollo/client';`

Akirtovskis: @pradel but pipelines are failing now 😬 

Actually if I run yarn generate from `client` it automatically updates to :

```
import * as ApolloReactCommon from '@apollo/react-common';
import * as ApolloReactHooks from '@apollo/react-hooks';
```

pradel: my bad I didn't see it was from the generated file 🤦

pradel (approved):

PR #130, opened 2020-09-24
feat: unlinking db from app functionality

Akirtovskis:


client/src/index.tsx
```diff
@@ -25,10 +27,17 @@ const GlobalStyle = createGlobalStyle`
   }
 `;
 
-const httpLink = createHttpLink({
+const httpLink = new HttpLink({
   uri: `${config.serverUrl}/graphql`,
 });
 
+const wsLink = new WebSocketLink({
+  uri: `ws://localhost:4000/graphql`,
```
pradel, line 35: if we harcode the value it won't work when deployed to an external server (on prod)

pradel: Maybe we can use `config.serverUrl` but we need to replace the HTTP with ws

client/src/index.tsx
```diff
@@ -53,14 +62,26 @@ const errorLink = onError(({ graphQLErrors }) => {
...
+    return (
+      definition.kind === 'OperationDefinition' &&
+      definition.operation === 'subscription'
+    );
+  },
+  wsLink,
+  httpLink
+);
+
+const client = new ApolloClient({
```
pradel, line 77: can we keep the name `apolloClient`?

server/src/queues/unlinkDatabase.ts
```diff
@@ -0,0 +1,80 @@
...
+ * - Unlink db
+ */
+const worker = new Worker(
+  queueName,
+  async (job) => {
+    const { app, database } = job.data;
+    debug(
+      `starting unlinkDatabaseQueue for ${database.type} database ${database.name} from  ${app.name} app`
+    );
+    console.log(
```
pradel, line 37: let's remove all the console.log before merging :)

server/src/queues/unlinkDatabase.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const ssh = await sshConnect();
+
+    const res = await ssh.execCommand(
+      `${dbType}:unlink ${database.name} ${app.name}`
+    );
+
+    const arrayOfLogs = res.stdout.split('\n');
+
+    pubsub.publish('DATABASE_UNLINKED', { unlinkDatabaseLogs: arrayOfLogs });
```
pradel, line 51: the problem of sending the logs at the end of the exec command is that if the task is taking 10 mins, we will only get the log in the UI 10 mins after. We should rather stream them to have them in real time.

server/src/index.ts
```diff
@@ -219,6 +246,8 @@ const apolloServer = new ApolloServer({
 });
 apolloServer.applyMiddleware({ app });
 
+apolloServer.installSubscriptionHandlers(http);
```
pradel, line 249: Do you think we should remove socket.io in this pr or in another one?

server/src/graphql/mutations/unlinkDatabase.ts
```diff
@@ -0,0 +1,68 @@
...
+      },
+      include: {
+        apps: {
+          where: {
+            id: appId,
+          },
+        },
+      },
+    }),
+    prisma.app.findOne({
```
pradel, line 31: do you need to do this find as the app is already included in the previous query?

server/src/graphql/mutations/unlinkDatabase.ts
```diff
@@ -0,0 +1,68 @@
...
+  const isLinked = database.apps.length === 1;
+
+  if (!isLinked) {
+    throw new Error(
+      `${database.name} database is not linked to ${app.name} app`
+    );
+  }
+
+  await unlinkDatabaseQueue.add('unlink-database', {
+    app,
```
pradel, line 63: It's usually better to pass only the id when working with queues (and then do a findOne at the beginning of the queue to retrieve these objects). As the object can be pretty big and exceed the max size allowed.

pradel (changes requested):
First review (mostly server), I will try the UI soon :)

client/src/index.tsx
pradel: Actually, this won't work `client/src/config.ts`, we need to find a good way to make in work on prod

server/src/graphql/mutations/unlinkDatabase.ts
Akirtovskis: @pradel fixed.

server/src/graphql/mutations/unlinkDatabase.ts
Akirtovskis: @pradel no, not really. Fixed!

server/src/index.ts
Akirtovskis: @pradel I am for doing this in a separate PR. Created and issue for this, so we don't forget : https://github.com/ledokku/ledokku/issues/133 .

server/src/queues/unlinkDatabase.ts
Akirtovskis: @pradel yeah, this one I was not really able to figure out and currently we are sending one bulk of text 😬 .

server/src/queues/unlinkDatabase.ts
Akirtovskis: @pradel done.

client/src/index.tsx
Akirtovskis: Resolving this here as new issue is already opened.

pradel (approved):
🚀

PR #131, opened 2020-09-25
feat(client): improve app logs view

pradel:
Show the same colors as in the terminal, that way errors will be in red etc 😍 

When there are no logs
Screenshot 2020-09-25 at 18 44 30

New view with colors
Screenshot 2020-09-25 at 19 10 00


Akirtovskis (approved):
🚀

PR #145, opened 2020-10-01
feat(client): improve modal component

Akirtovskis:
@pradel 
- modal is now more flexible than before
- terminal is now beautifully fitting in in all of the unlinking/linking operations
- some warnings fixed
- found a bug within unlinking and fixed it 

there are still UI things that within modal that can be made better, but for MVP this looks solid enough

closes : https://github.com/ledokku/ledokku/issues/135

### DIFF
--- a/client/src/ui/components/Modal.tsx
+++ b/client/src/ui/components/Modal.tsx
@@ -6,17 +6,12 @@ interface ModalProps {
   className?: string;
 }
 
-export const Modal = ({ children, className, ...props }: ModalProps) => (
-  <React.Fragment>
-    <div
-      {...props}
-      className="bg-gray-400 bg-opacity-25 absolute inset-0 flex justify-center items-center"
-    >
-      <div className="bg-white rounded-md shadow-xl w-156 py-6 px-6">
-        {children}
-      </div>
+export const Modal = ({ children, className }: ModalProps) => (
+  <div className="bg-gray-400 bg-opacity-25 absolute inset-0 flex justify-center items-center">
+    <div className="bg-white rounded-md shadow-xl w-156 py-6 px-6">
+      {children}
     </div>
-  </React.Fragment>
+  </div>
 );
 
 interface ModalTitleProps {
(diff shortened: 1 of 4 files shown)


client/src/ui/components/Modal.tsx
```diff
@@ -1,99 +1,89 @@
...
-  mainText: React.ReactNode;
-  header: string;
-  closeModal: () => void;
-  ctaFn: () => void;
+  children?: React.ReactNode;
+  className?: string;
+}
+
+export const Modal = ({ children, className, ...props }: ModalProps) => (
+  <React.Fragment>
```
pradel, line 10: is the fragment really useful here?

client/src/ui/components/Modal.tsx
```diff
@@ -1,99 +1,89 @@
...
+    <div
+      {...props}
+      className="bg-gray-400 bg-opacity-25 absolute inset-0 flex justify-center items-center"
```
pradel, line 13: the classnames are not merged so we can't extend it

client/src/ui/components/Modal.tsx
```diff
@@ -1,99 +1,89 @@
...
+  children?: React.ReactNode;
+  className?: string;
+}
+
+export const ModalTitle = ({
+  children,
+  className,
+  ...props
+}: ModalTitleProps) => (
+  <div {...props} className="flex justify-between items-center">
```
pradel, line 32: the classnames are not merged so we can't extend it

Akirtovskis:
@pradel all feedbacks fixed 🚀

client/src/ui/components/Modal.tsx
```diff
@@ -1,99 +1,71 @@
...
+
+export const ModalTitle = ({ children, className }: ModalTitleProps) => (
+  <div className="flex justify-between items-center">
+    <h3 className="text-lg font-medium">{children}</h3>
+  </div>
+);
+
+interface ModalDescriptionProps {
+  children?: React.ReactNode;
+  className?: string;
```
pradel, line 30: @Akirtovskis className is still allowed but if you pass one currently it will not work. You will need to merge this prop with the current className

Akirtovskis: @pradel should be good now after the latest commit.

pradel (approved):

Issue #155, opened 2020-10-04
Create queue for database creation process


pradel:
# Feature request

- create a queue
- display the creation logs in realtime in the ui

PR #172, opened 2020-10-06
feat: show real time logs on db creation

Akirtovskis:
@pradel 

Updated flow description : 
- once we click create, terminal appears 
- logs are rendering 
- depending on the end message after the process is done  we see either `next` or `back` button 
- depending on the end message after the process is done we see either  `toast success` or `toast error`
- if we click `next` it takes us to the newly created db page 
- if we click `back` it takes us back to database creation page

Preview screenshot :
image

Closes : https://github.com/ledokku/ledokku/issues/155


client/src/pages/create-database.tsx
```diff
@@ -104,118 +126,211 @@ export const CreateDatabase = () => {
 
   const isPluginInstalled = data?.isPluginInstalled.isPluginInstalled;
 
+  const handleNext = () => {
+    try {
+      subscribeToMore({
```
Akirtovskis, line 131: @pradel this works, but would like to hear some feedback whether it's a good solution. From what I understand it subscribes to query and updates the result if there's any update. 

Is that correct and does application of this `subscribeToMore` make sense in this case?

pradel: @Akirtovskis I am not sure what you are tried to achieve here. If you are just trying to get the database id why don't you send it from the server via the `end` event?

Akirtovskis: @pradel I was indeed just trying to get the id of db and thanks for the elegant & simple suggestion.  Fixed now.

server/src/queues/createDatabase.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+    });
+
+    debug(
+      `finishing createDatabase queue for ${dbType} database called ${databaseName}`
+    );
+    if (!res.stderr) {
+      pubsub.publish('DATABASE_CREATED', {
+        createDatabaseLogs: {
+          message: `Successfully created DB with ID: ${createdDb.id}`,
```
pradel, line 78: I think this would be cleaner if we only pass the id, the text can be done in the frontend only. That way you don't have to parse the message client side.

Also, what do you think about separating the end type to know if it was a failure or success via the type, eg: `end-success`, `end-failure` or `end:success`, `end:failure`?

server/src/lib/dokku/plugin/database/create.ts
```diff
@@ -1,15 +0,0 @@
-import { NodeSSH } from 'node-ssh';
-
-export const create = async (
-  ssh: NodeSSH,
-  name: string,
-  databaseType: string
-) => {
```
pradel, line 7: @Akirtovskis what do you think about keeping this file but accepting a 4th optional argument that will be the options passed down to `execCommand`? That way the command is still abstracted, so easier to reuse if we need later :)

client/src/pages/dashboard.tsx
```diff
@@ -11,13 +11,17 @@ import { MySQLIcon } from '../ui/icons/MySQLIcon';
...
   // const history = useHistory();
-  const { data /* loading, error */ } = useDashboardQuery({});
+  const { data, refetch /* loading, error */ } = useDashboardQuery({});
 
   // TODO show loading
   // TODO handle error
 
   // TODO if no apps or dbs show onboarding screen
 
+  useEffect(() => {
```
pradel, line 21: not a big fan of this, as it can leads to some weird behavior. Are you trying to fetch the latest data from the server when you navigate to this page?
If yes it's cleaner to change the fetch policy of the query to cache and network.

client/src/pages/create-database.tsx
```diff
@@ -104,118 +130,207 @@ export const CreateDatabase = () => {
...
         <h1 className="text-lg font-bold">Create a new database</h1>
+        <div className="mt-12">
+          {isTerminalVisible ? (
+            <React.Fragment>
+              <p className="mb-2 ">
+                Creating <b>{formik.values.type}</b> database{' '}
+                <b>{formik.values.name}</b>
+              </p>
+              <Terminal className={'w-6/6'}>
+                <p className="text-green-400 mb-2">
```
pradel, line 179: in my personal opinion text inside the terminal should come from the server only. What do you think about moving this sentence in top of the terminal? Maybe in some light grey color?

server/src/queues/createDatabase.ts
Akirtovskis: @pradel different types for end messages totally make sense!

client/src/pages/dashboard.tsx
Akirtovskis: @pradel yes, the intended behaviour is to get the latest data. Thanks for the suggestion, on it.

server/src/lib/dokku/plugin/database/create.ts
Akirtovskis: @pradel makes sense and probably worth to do the same for link and unlink as well. For this fn I will do in this PR, but for others create another to not make a mess out of it.

pradel: Sounds good 👍

pradel (approved):

PR #180, opened 2020-10-08
feat(server): add extra validation for subscription connection

Akirtovskis:


server/src/index.ts
```diff
@@ -234,13 +235,27 @@ const apolloServer = new ApolloServer({
...
-        // TODO ARTURS : FIND USER FN FOR EXTRA LAYER OF SECURITY
+        const decoded = jsonwebtoken.verify(
+          context.token,
+          config.jwtSecret
+        ) as {
+          userId: string;
+        };
+        const userId = decoded.userId;
+        try {
+          await prisma.user.findOne({
```
pradel, line 251: @Akirtovskis I think the function will return null if the user is not found instead of throwing no?

Akirtovskis: @pradel you are right, fixed.

pradel (approved):

PR #183, opened 2020-10-13
chore: add name validation on db/app creation

Akirtovskis:
can be merged only after : https://github.com/ledokku/ledokku/pull/172

@pradel 
- this pr adds name validation for both creating db and creating app 
- for the app db it should be 💯 good to go
- for db creation I am struggling with the validation for `one name per db type` as I imagine we wouldn't want to limit name `to one db name per server` so I added a comment below to point out the problematic part


client/src/pages/create-database.tsx
```diff
@@ -63,10 +64,67 @@ const DatabaseBox = ({ label, selected, icon, onClick }: DatabaseBoxProps) => {
...
+      .string()
+      .oneOf(['POSTGRESQL', 'MYSQL', 'MONGODB', 'REDIS'])
+      .required(),
+    name: yup
+      .string()
+      .required('Database name is required')
+      .matches(/^[a-z0-9-]+$/)
+      .when(['type'], {
+        is: (val) => val === 'POSTGRESQL',
+        // val === databaseQueryData?.databases.find((db) => db.type === val),
```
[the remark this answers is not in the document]
pradel: I guess the best to check if the name is free is to actually make a new request to the server every time the user changes the name (with some denounce of course).

client/src/pages/create-app.tsx
```diff
@@ -48,27 +69,37 @@ export const CreateApp = () => {
...
-            >
-              Create
-            </Button>
+        <form>
+          <div className="grid grid-cols-1 md:grid-cols-3">
+            <div>
+              <FormLabel>App name: </FormLabel>
+              <FormInput
+                autoComplete="off"
+                // className="inline w-full max-w-xs bg-white border border-grey rounded py-3 px-3 text-sm leading-tight transition duration-200 focus:outline-none focus:border-black"
```
pradel, line 78: @Akirtovskis let's remove this before merging?

client/src/pages/create-app.tsx
```diff
@@ -1,18 +1,39 @@
...
+
+  const createAppSchema = yup.object().shape({
+    name: yup
+      .string()
+      .required('App name is required')
+      .matches(/^[a-z0-9-]+$/)
+      .test(
+        'Name exists',
+        'App with this name already exists',
+        (val) => !data?.apps.find((app) => app.name === val)
```
pradel, line 27: @Akirtovskis we should also check the type of the db with the name I think

client/src/pages/create-app.tsx
```diff
@@ -1,18 +1,39 @@
...
   const formik = useFormik<{ name: string }>({
     initialValues: {
       name: '',
     },
+
+    validateOnChange: true,
+    validationSchema: createAppSchema,
     // TODO yup validation
     onSubmit: async (values) => {
       // TODO validate name
```
pradel, line 39: you can remove this TODO

client/src/pages/create-database.tsx
```diff
@@ -203,7 +222,11 @@ export const CreateDatabase = () => {
               ) : null}
             </React.Fragment>
           ) : (
-            <form onSubmit={formik.handleSubmit} className="mt-8">
+            <form
+              onSubmit={formik.handleSubmit}
+              onChange={formik.handleChange}
```
pradel, line 227: Hum I am not sure about this onChange 🤔 I never saw that before

server/src/graphql/mutations/createApp.ts
```diff
@@ -14,6 +14,14 @@ export const createApp: MutationResolvers['createApp'] = async (
     throw new Error('Unauthorized');
   }
 
+  const apps = await prisma.app.findMany();
```
pradel, line 17: would be better to do a `findMany` with the name directly and check the item 0 in the array if it's set

pradel: This step should be done after the yup validation :)

server/src/graphql/mutations/createDatabase.ts
```diff
@@ -25,6 +25,15 @@ export const createDatabase: MutationResolvers['createDatabase'] = async (
     throw new Error('Unauthorized');
   }
 
+  const databases = await prisma.database.findMany();
```
pradel, line 28: same comments as server/src/graphql/mutations/createApp.ts :)

server/src/graphql/mutations/createDatabase.ts
```diff
@@ -28,6 +28,18 @@ export const createDatabase: MutationResolvers['createDatabase'] = async (
   // We make sure the name is valid to avoid security risks
   createDatabaseSchema.validateSync({ name: input.name });
 
+  const databases = await prisma.database.findMany({
+    where: {
+      name: input.name,
```
pradel, line 33: @Akirtovskis you can also add the type in the where condition that way you don't need the find afterwards :)

Akirtovskis: @pradel Done! 🙌

pradel (approved):
🚀

Issue #184, opened 2020-10-18
Recreate the unlink/link functions in server/dokku/lib for later use


Akirtovskis:
# Bug report

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository.

## System information

- Version of dokku: [e.g. 0.21.4]
- Version of ledokku: [e.g. 0.0.1]

PR #200, opened 2020-10-20
feat(server): recreate link and unlink functions in dokku lib

Akirtovskis:
Closes : https://github.com/ledokku/ledokku/issues/184

And this is also 200th PR 🚀

### DIFF
--- a/server/src/lib/dokku/plugin/database/link.ts
+++ b/server/src/lib/dokku/plugin/database/link.ts
@@ -1,0 +1,16 @@
+import { NodeSSH, SSHExecOptions } from 'node-ssh';
+
+export const link = async (
+  ssh: NodeSSH,
+  databaseName: string,
+  databaseType: string,
+  appName: string,
+  options: SSHExecOptions
+) => {
+  const resultDatabaseLink = await ssh.execCommand(
+    `${databaseType}:link ${databaseName} ${appName}`,
+    options
+  );
+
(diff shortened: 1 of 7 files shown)


server/src/lib/dokku/plugin/database/link.ts
```diff
@@ -0,0 +1,16 @@
+import { NodeSSH, SSHExecOptions } from 'node-ssh';
+
+export const link = async (
+  ssh: NodeSSH,
+  databaseName: string,
+  databaseType: string,
+  appName: string,
+  options: SSHExecOptions
```
pradel, line 8: option should be optional I think :)

server/src/lib/dokku/plugin/database/unlink.ts
```diff
@@ -0,0 +1,16 @@
+import { NodeSSH, SSHExecOptions } from 'node-ssh';
+
+export const unlink = async (
+  ssh: NodeSSH,
+  databaseName: string,
+  databaseType: string,
+  appName: string,
+  options: SSHExecOptions
```
pradel, line 8: option should be optional I think :)

server/src/lib/dokku/plugin/database/link.ts
Akirtovskis: @pradel you are right, fixed both!

pradel (approved):

PR #205, opened 2020-10-22
feat(bootstrap): add rule that allows access port 4000

Akirtovskis:


ledokku-bootstrap.sh
```diff
@@ -87,6 +87,9 @@ main() {
   docker tag ledokku/ledokku:0.3.0 dokku/ledokku:0.3.0
   dokku tags:deploy ledokku 0.3.0
 
+  # We setup a rule that allows access to port 4000
+  ufw allow 4000
```
pradel, line 91: @Akirtovskis this really depends on the server, we should not do this for the user as they might not use ufw. What we could do is to add it as a suggestion after `echo "http://${DOKKU_SSH_HOST}:4000"`

pradel (approved):

PR #213, opened 2020-10-23
chore(client): switch from Select to Listbox

Akirtovskis:
@pradel thoughts on this? 

If it looks good and code is ok, then I will add the same for the other unlink section. 

Also does it make sense to create a separate component for this in `ui` folder?

image


pradel:
@Akirtovskis maybe we should wait to have another totally different usage of the listbox component, that way we could really abstract the component in a good way more easily.
But if you want to give it a try in this pr, feel free :)

Issue #168, opened 2020-10-06
Use new headless UI ListBox component for link/unlink selects


Akirtovskis:
# Feature request

## Is your feature request related to a problem? Please describe.

See : https://github.com/tailwindlabs/headlessui/tree/develop/packages/%40headlessui-react#listbox-select

PR #215, opened 2020-10-24
chore(client): switch dropdown from react-select to headless UI

Akirtovskis:
Closes : https://github.com/ledokku/ledokku/issues/168

image

### DIFF
--- a/client/src/pages/app/index.tsx
+++ b/client/src/pages/app/index.tsx
@@ -303,6 +303,7 @@ export const App = () => {
                       color="grey"
                       width="large"
                       className="mt-2"
+                      isDisabledBelowDropdown={true}
                       isLoading={
                         databaseLinkLoading &&
                         !databaseLinkData &&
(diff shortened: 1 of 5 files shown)


client/src/ui/components/Button.tsx
```diff
@@ -17,6 +18,7 @@ export const Button = ({
   iconEnd,
   iconStart,
   isLoading,
+  isDisabledBelowDropdown,
```
Akirtovskis, line 21: @pradel I added this prop because when disabled button was with `opacity-50`, it was appearing in front of dropdown menu. 

For all the other disabled cases this opacity way works fine, but for the case below dropdown it didn't not so therefore the hacky way. 

Do you see any better way this could be fixed?

client/src/pages/app/index.tsx
```diff
@@ -257,6 +317,7 @@ export const App = () => {
                       color="grey"
                       width="large"
                       className="mt-2"
+                      // isDisabledBelowDropdown={true}
```
pradel, line 320: this can be removed

pradel (approved):
Good to go, if you can just remove the comment before merging, otherwise feel free to merge 🚀

Issue #196, opened 2020-10-20
Pre-populate the database name when creating a database from the app view


pradel:
# Feature request

When creating a database from the app view we can pre-generate the name of the database with the name of the app.

Eg: I am on the ledokku app view, I click to create a database, the field is already populated with "ledokku" on the create database page.

PR #235, opened 2020-11-08
feat(client): pre populate db name on creation

Akirtovskis:
closes : https://github.com/ledokku/ledokku/issues/196

### DIFF
--- a/client/src/pages/app/index.tsx
+++ b/client/src/pages/app/index.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { toast } from 'react-toastify';
-import { useHistory } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


client/src/pages/app/index.tsx
```diff
@@ -256,7 +261,10 @@ export const App = () => {
                       onChange={
                         selectedDb.value.name !== 'Create new database'
                           ? setSelectedDb
-                          : history.push('/create-database')
+                          : history.push({
+                              pathname: '/create-database',
+                              state: app.name,
```
pradel, line 266: just wondering, shouldn't state be an object?

Akirtovskis: it can be, but works w/o being object as well.

pradel (approved):

PR #238, opened 2020-11-15
feat(client): add restart app functionality

Akirtovskis:
New settings look screams a bit of redesign as the single column is packed, but that can be done in another PR.

image

### DIFF
--- a/client/src/pages/app/settings.tsx
+++ b/client/src/pages/app/settings.tsx
@@ -43,10 +43,7 @@ export const Settings = () => {
     { loading: destroyAppMutationLoading },
   ] = useDestroyAppMutation();
 
-  const [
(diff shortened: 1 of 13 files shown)


client/src/pages/app/settings.tsx
```diff
@@ -110,6 +168,68 @@ export const Settings = () => {
             </div>
 
             <AppProxyPorts appId={app.id} />
+            <h1 className="text-md font-bold mt-6">Restart app</h1>
```
pradel, line 171: should we split it in a separate component as "AppProxyPorts"?

pradel (commented):
One small question otherwise the rest is good 👍

pradel (approved):

Issue #198, opened 2020-10-20
Domain management


pradel:
# Feature request

Allow user to manage the domain names (bonus point https) via the UI.

> 1) attach subdomains to dokku app: dokku domains:add app domain.app.com
> 2) install official letsencrypt dokku plugin and run one command

PR #252, opened 2020-11-29
feat(client): app domain managment

Akirtovskis:
closes : https://github.com/ledokku/ledokku/issues/198

`setDomain` -> overwrites the domain list with the domain that is provided to fn
`addDomaiun` -> adds domain to the list of existing domains

I left both of methods in as I suspect we might need to use this later, but for now only add is used. 

Screenshot of new looks below. 

image


pradel (approved):
Really cool!

pradel (approved):
Really cool!

PR #265, opened 2021-02-08
feat(client,server): deploy directly from public git repo

Akirtovskis:
TODO : 
- [x] adding githubRepoId to app in DB
- [x] adding env var setup to verify webhook fn
- [x] info UI about app creation
- [x] app creation from gh repo
- [x] webhook initial setup
- [x] securing webhook requests
- [x] docs about webhooks setup
- [x] add choose source step
- [x] split appCreation and appGithubCreation 

Adds functionality for deploying apps straight outta GH from public repo. 

Create app view
image

App view after freshly created app with git url
image

Choosing source view 
image


client/src/pages/create-app.tsx
```diff
@@ -23,81 +55,205 @@ export const CreateApp = () => {
...
+                breathe out, logs are about to appear below:
+              </p>
+              <Terminal className={'w-6/6'}>
+                {console.log(arrayOfCreateAppLogs)}
+                {arrayOfCreateAppLogs.map((log) => (
+                  <p
+                    key={arrayOfCreateAppLogs.indexOf(log)}
+                    className={'text-s leading-5'}
+                  >
+                    {log.message?.replaceAll('[1G', '')}
```
Akirtovskis, line 137: @pradel this `replaceAll` thing seems pretty bad, but it works. Using the same on the server side before sending the log somehow was not working.

pradel:
@Akirtovskis do you have a screenshot for this?

nerdsupremacist:
This seems like a cool idea. I have a few suggestions:

1. Allow the user to select from any of their repos (we have access to their github already). And at any time, not only during creation
2. And I would suggest doing this via a webhook: that way it should work with private repos too.
- So we would have to add an endpoint that would trigger the deployment
- And setting it up would add this endpoint as a webhook in the repository

Akirtovskis:
Hey @nerdsupremacist , thanks for the feedback and reaching out :) 

As far as we currently have tried, public repos will work with what you proposed and our idea to implement that was very similar to that However with private repos it gets much trickier.

- With our login scope we do not have access to private repos of said user
- Currently to build from a private repo via Dokku you would need to enter GH personal access token, which is kind of a master key for all your GH details and therefore something that we do not want to have a hold on for now or possibly ever. 
- Next Dokku version will include a fix which will enable us to build an app with GH username and password which seems something that we can work with. Link to related issue : https://github.com/dokku/dokku/issues/4364 

If there's something obvious that I am missing, would be curious to hear. 

Also @pradel if you have more comments on this feel free to jump in.

pradel:
Just want to add that it looks like the only way to have a token that can only access a few selected repositories (and limited rights for the repo of course) would be to create a Github Application, that way we could also automatically create the webhooks for the user. But this would require much more work so it might be something we want to investigate later once this functionality is already working.

server/src/lib/webhooks/webhooks.ts
```diff
@@ -0,0 +1,22 @@
...
+    where: {
+      githubRepoId: req.body.repository.id.toString(),
+    },
+  });
+
+  const branch = req.body.ref.replace('refs/heads/', '');
+  const { name, git_url } = req.body.repository;
+
+  if (appToRedeploy) {
+    await dokku.git.sync(ssh, name, git_url, branch);
```
pradel, line 20: this should probably be done in a queue to not block the request

server/src/queues/createApp.ts
```diff
@@ -0,0 +1,138 @@
...
+    debug(
+      `starting create app queue for ${userId} and ${appName} app with repoURL ${gitRepoUrl} and branch name ${branchName}`
+    );
+
+    const branch = branchName ? branchName : 'main';
+
+    const ssh = await sshConnect();
+
+    //to do move this out to UTILS
+    const getRepoData = (gitRepoUrl: string) => {
```
pradel, line 47: lets move this logic in the mutation, the validation should be done before the queue start in case the user send a non existing repo for example

client/src/modules/webhooks/Webhooks.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { app } = data;
+
+  if (!app) {
+    // TODO nice 404
+    return <p>App not found.</p>;
+  }
+
+  return (
+    <>
+      <Heading size="sm" className="text-md font-bold pb-1 py-5">
```
pradel, line 41: can you remove the className and use chakra values?

client/src/modules/webhooks/Webhooks.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <Grid templateColumns="repeat(5, 1fr)" gap={1}>
+        <GridItem colSpan={2} pt={3}>
+          <Text>
+            {isSecretVisible
+              ? data?.app?.githubRepoId + config.githubClientId.slice(0, 4)
+              : '*************'}
+          </Text>
+        </GridItem>
+        <GridItem colSpan={2} ml={-6}>
+          <svg
```
pradel, line 57: can you remove the className and use chakra values?

you can use `chakra.svg` to use the custom props

server/src/graphql/mutations/createApp.ts
```diff
@@ -28,17 +31,56 @@ export const createApp: MutationResolvers['createApp'] = async (
   if (isAppNameTaken) {
     throw new Error('App name already taken');
   }
+  // we do git related operations only for apps
+  // that have gitRepoURl
+  let repo;
+  if (input.gitRepoUrl) {
+    const repoData = getRepoData(input.gitRepoUrl);
+    try {
+      const res = await fetch(
```
pradel, line 40: you can use the `octokit/rest` package that will give you correct ts typings :)

server/src/graphql/mutations/createApp.ts
```diff
@@ -28,17 +31,56 @@ export const createApp: MutationResolvers['createApp'] = async (
...
+        `https://api.github.com/repos/${repoData.owner}/${repoData.repoName}`
+      );
+      repo = await res.json();
+    } catch (error) {
+      console.log(error);
+    }
+  }
```
pradel, line 47: Should we add a check to verify that the user own the repo for now? You should not be allowed to clone one of my public repo I think.

server/src/index.ts
```diff
@@ -404,6 +414,16 @@ app.get('*', (_, res) => {
   );
 });
 
+app.post('/webhooks', (req, res) => {
+  const isWebhookVerified = verifyWebhookSecret(req);
+  if (!isWebhookVerified) {
+    res.status(400).send('Request not verified');
```
pradel, line 420: let's use the debug package here so if some people have an issue we can see what's going on

server/src/lib/dokku/git/sync.ts
```diff
@@ -0,0 +1,15 @@
+import { NodeSSH, SSHExecOptions } from 'node-ssh';
+
+export const sync = async (
+  ssh: NodeSSH,
+  appName: string,
```
pradel, line 5: can we use named arguments here? It's fairly easy to do a mistake when there are a lot of args otherwise

server/src/lib/webhooks/utils.ts
```diff
@@ -0,0 +1,29 @@
+import crypto from 'crypto';
+import { Request } from 'express';
+
+export const verifyWebhookSecret = (req: Request) => {
+  // we get signature from gh
+  const signature = req.header('X-Hub-Signature-256');
+  // we encrypt signature from our db
+
+  // we take repo id
+  const repoId = req.body.repository.id.toString();
```
pradel, line 10: if body is undefined this will create an error

server/src/lib/webhooks/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  // we take first 4 chars from gh client id
+  const clientIDPart = process.env.GITHUB_CLIENT_ID.slice(0, 4);
```
pradel, line 13: I had some thought about this one, and if the secret of one repo leak you can guess the other secrets which is not good..
What do you think about adding a new field to the app database schema and generate a random secret during the app creation? Then we can show it to the user in the settings

server/src/lib/webhooks/webhooks.ts
```diff
@@ -0,0 +1,22 @@
...
+import { deployAppQueue } from './../../queues/deployApp';
+
+export const githubPushWebhookHandler = async (req: Request) => {
+  const appToRedeploy = await prisma.app.findFirst({
+    where: {
+      githubRepoId: req.body.repository.id.toString(),
+    },
+  });
+
+  const branch = req.body.ref.replace('refs/heads/', '');
```
pradel, line 12: I think we have to save the branch in the database, you will receive an event for every push on any branch (so even on feature ones for example). You only want to deploy for the one selected by the user.

server/src/server.ts
```diff
@@ -2,4 +2,5 @@ import express from 'express';
 import { createServer } from 'http';
 
 export const app = express();
+app.use(express.json());
```
pradel, line 5: make more sense to have this in index.ts around line 18

server/src/graphql/mutations/createApp.ts
Akirtovskis: @pradel for this one I do not think that there's any benefit for having that check. 

Multiple cases come in mind for this : 
- just want to test with some generic public repo w/o forking 
- login from one github acc and want to test something from some gh org that is  public

server/src/lib/dokku/git/sync.ts
Akirtovskis: Sure, will add.

server/src/index.ts
Akirtovskis: Sounds good, will do.

server/src/lib/webhooks/utils.ts
Akirtovskis: Yeah, this sounds best.

server/src/server.ts
Akirtovskis: Resolved.

server/src/graphql/mutations/createApp.ts
Akirtovskis: Resolved as this will be done when fixing the : https://github.com/ledokku/ledokku/issues/295

server/src/graphql/mutations/createApp.ts
Akirtovskis: Resolved.

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (isAppNameTaken) {
+    throw new Error('App name already taken');
+  }
+
+  const octokit = new Octokit({});
+
+  let repo;
+
+  const repoData = getRepoData(input.gitRepoUrl);
+  try {
```
pradel, line 40: is the try catch needed? What is happening if repo does not exist? Is it throwing or returning null?

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+  } catch (error) {
+    console.log(error);
+  }
+
+  const ssh = await sshConnect();
+
+  const dokkuApp = await dokku.apps.create(ssh, input.name);
+
+  const hash = crypto
```
pradel, line 53: a simple way to generate a random token
```
import { randomBytes } from 'crypto';

const generateRandomToken = (length = 43): string => randomBytes(length).toString('hex');
```

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -0,0 +1,111 @@
...
+  _,
+  { input },
+  { userId }
+) => {
+  if (!userId) {
+    throw new Error('Unauthorized');
+  }
+
+  // We make sure the name is valid to avoid security risks
+  appNameSchema.validateSync({ name: input.name });
```
pradel, line 21: I guess we should also validate that the URL is a valid url

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -0,0 +1,111 @@
...
+    throw new Error('App name already taken');
+  }
+
+  const octokit = new Octokit({});
+
+  let repo;
+
+  const repoData = getRepoData(input.gitRepoUrl);
+  try {
+    repo = await octokit.repos.get({
```
pradel, line 41: here we should that the repo exist but we should also check that the branch exist for that repo otherwise it could create issues later

server/src/graphql/mutations/destroyApp.ts
```diff
@@ -54,6 +54,22 @@ export const destroyApp: MutationResolvers['destroyApp'] = async (
...
+  const githubMetaToDelete = await prisma.app
+    .findUnique({
+      where: {
+        id: appId,
+      },
+    })
+    .AppMetaGithub();
+
+  if (githubMetaToDelete) {
+    await prisma.appMetaGithub.delete({
```
pradel, line 66: is there a way to do cascading deletion in prisma?

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -0,0 +1,111 @@
...
+    .slice(0, 40);
+
+  const app = await prisma.app.create({
+    data: {
+      name: input.name,
+      type: 'GITHUB',
+    },
+  });
+
+  await prisma.appMetaGithub.create({
```
pradel, line 66: this can be merged in one operation with prisma create line 59

server/src/index.ts
```diff
@@ -247,6 +278,7 @@ const typeDefs = gql`
   type Query {
     setup: SetupResult!
     apps: [App!]!
+    appMetaGithub(appId: String!): AppMetaGithub
```
pradel, line 281: as we can access this via a subfield of app is it really needed?

server/src/lib/webhooks/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+import crypto from 'crypto';
+import { Request } from 'express';
+
+export const verifyWebhookSecret = async (req: Request) => {
+  // we get signature from gh
+  const signature = req.header('X-Hub-Signature-256');
+
+  const appToRedeploy = await prisma.appMetaGithub.findFirst({
+    where: {
+      repoId: req.body.repository.id.toString(),
```
pradel, line 11: if req.body is null this will crash

server/src/queues/deployApp.ts
```diff
@@ -0,0 +1,104 @@
...
+const queueName = 'deploy-app';
+const debug = createDebug(`queue:${queueName}`);
+const redisClient = new Redis(config.redisUrl);
+
+interface QueueArgs {
+  appName: string;
+  gitRepoUrl: string;
+  branchName?: string;
+  dokkuApp?: boolean;
+  appId?: string;
```
pradel, line 18: let's just take the `appId` as argument as we can fetch the rest of the data from the db

server/src/queues/deployApp.ts
```diff
@@ -0,0 +1,104 @@
...
+      `starting deploy app queue for ${appName} app with repoURL ${gitRepoUrl} and branch name ${branchName}`
+    );
+
+    const branch = branchName ? branchName : 'main';
+
+    const ssh = await sshConnect();
+
+    await dokku.git.unlock(ssh, appName);
+
+    if (gitRepoUrl) {
```
pradel, line 47: is this if needed?

client/src/pages/app-creation/choose-source.tsx
```diff
@@ -0,0 +1,181 @@
...
+                    onClick={() => formik.setFieldValue('type', 'GITHUB')}
+                  />
+                  <SourceBox
+                    selected={formik.values.type === AppTypes.DOKKU}
+                    label="Dokku"
+                    icon={
+                      <Image
+                        boxSize="48px"
+                        objectFit="cover"
+                        src="https://alternative.me/media/256/dokku-icon-2rtcs9lztuk6iud7-c.png"
```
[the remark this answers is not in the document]
Akirtovskis: @pradel for this one I just added img to our public folder to be safe for now. If later on we find good svg, can be changed to chakra.svg.

pradel (approved):

server/src/graphql/mutations/destroyApp.ts
Akirtovskis: @pradel the issue mentioned above is still active and doesn't seem to be cleanly fixed.

PR #269, opened 2021-02-09
chore(client): partly migrate app views to chakra

Akirtovskis:
Migrated the header part for all screens and also table part for app index view. 

App main page screenshot : 
image


pradel (approved):
Nice!

PR #272, opened 2021-02-25
chore(docs): update version and links

Akirtovskis:
Update links to new dokku domain and latest version in bootstrap script


pradel (approved):
we will need to think of a way to automate this every time a new release is published to avoid having to do it manually

PR #278, opened 2021-02-28
chore: migrate create database view to chakra

pradel:


client/src/pages/create-database.tsx
```diff
@@ -93,27 +102,27 @@ export const CreateDatabase = () => {
...
-            'Name already exists',
-            `You already have created ${type} database with this name`,
-            (name) =>
-              !dataDb?.databases.find(
-                (db) => db.name === name && type === db.type
-              )
-          );
-      }),
+      .matches(/^[a-z0-9-]+$/),
+    // .when('type', (type: DatabaseTypes) => {
```
pradel, line 114: @Akirtovskis this part is actually breaking the validation, by commenting it we check that the field is not empty and match the regex. Do you have an idea why by any chance?

pradel:
Except for the validation issue (not related to this pr originally but can be good to fix at the same time), this is good to go

client/src/pages/create-database.tsx
Akirtovskis: @pradel not immediate idea, would have to try this on fresh branch. Let's merge this PR with this part commented out and create an issue to return to this later?

Akirtovskis: @pradel not immediate idea, will check now on a fresh branch.

Akirtovskis: @pradel I will create issue as this seems like an issue multiple people are facing here : https://github.com/jquense/yup/issues/851 . Definitely there should be a way around, but let's leave this part commented out for now.

Akirtovskis: This worked : 

```
 name: yup.string().when('type', (type: DatabaseTypes) => {
      return yup
        .string()
        .required('Database name is required')
        .matches(/^[a-z0-9-]+$/)
        .test(
          'Name already exists',
          `You already have created ${type} database with this name`,
          (name) =>
            !dataDb?.databases.find(
              (db) => db.name === name && type === db.type
            )
        );
    }),
```

PR #281, opened 2021-03-03
fix(client): fix create app and database button

quantumsheep:
There was an issue with the create button for apps, it was triggering the form's default behavior which resulted in a page reload.

The PR also includes a fix to prevent issues with the databases create button.


Akirtovskis (approved):
@quantumsheep great catch! Thanks for submitting a PR for this! 🚀

PR #282, opened 2021-03-03
chore: switch from yarn to npm

quantumsheep:
It's only a proposal, yarn have the same functionalities. However npm is installed by default which improves contributors experience and ease to help the project.


pradel:
@quantumsheep thanks for the pr!
I don't think we want to migrate to npm as we are using the workspace feature of yarn to manage our monorepo :)
yarn is commonly used in the javascript ecosystem by a lot of projects and has proven to be reliable and is usually faster than npm

quantumsheep:
@pradel fair enough :)

PR #283, opened 2021-03-03
chore(client): migrate toast messages to chakra UI

Akirtovskis:
closes : https://github.com/ledokku/ledokku/issues/266

Migrated all toasts to Chakra Ui 

In utils added config file that is reusable so in components we have to just add description. 

Screenshots below

image
image

### DIFF
--- a/client/src/modules/app/AppRebuild.tsx
+++ b/client/src/modules/app/AppRebuild.tsx
@@ -1,5 +1,4 @@
 import { useState } from 'react';
-import { useToast } from '@chakra-ui/toast';
 import {
   useAppByIdQuery,
   useRebuildAppMutation,
@@ -15,6 +14,7 @@ import {
   ModalButton,
   Terminal,
 } from '../../ui';
+import { useToast } from '../../ui/toast';
 
 interface AppRebuildProps {
   appId: string;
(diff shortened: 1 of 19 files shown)


client/src/modules/app/AppRebuild.tsx
```diff
@@ -1,11 +1,12 @@
 import { useState } from 'react';
-import { toast } from 'react-toastify';
+import { useToast } from '@chakra-ui/toast';
```
pradel, line 2: `usetoast` should be imported from `"@chakra-ui/react"`

client/src/modules/app/AppRebuild.tsx
```diff
@@ -69,7 +71,10 @@ export const AppRebuild = ({ appId }: AppRebuildProps) => {
       setIsTerminalVisible(true);
       setRebuildLoading(true);
     } catch (e) {
-      toast.error(e.message);
+      toast({
```
pradel, line 74: @Akirtovskis what do you think about creating our own component for the toast that extend from the chakra one?

I quite like the API of using `toast.error(e.message);` or `toast.success(e.message);` so if we can keep it that would be super cool :)

Akirtovskis: @pradel sounds good, will have a go at this. And definitely agree that toastify way looks neater.

pradel (approved):

PR #286, opened 2021-03-05
chore(installation): une new dokku git:from-image command

### LINKED ISSUE #284: Update ledokku installation and upgrade script
# Feature request

With the new release of dokku 0.24.0 we can now use the new `git:from-image` for the installation and probably upgrade of ledokku so the process will be simplified.

pradel:
I manually tested on a new DO droplet and it worked 🎉 

Closes #284


Akirtovskis (approved):
Lgtm 🚀

PR #294, opened 2021-03-11
feat: improve header style

pradel:
First version of the style. I will update the app/database info in the header in another pr.


Akirtovskis (approved):
🚀 🚀 🚀

Issue #300, opened 2021-03-12
Add ssh.dispose() to all the queues to not overwhelm server


Akirtovskis:
# Feature request

## Is your feature request related to a problem? Please describe.

A clear and concise description of what you want and what your use case is.

PR #306, opened 2021-03-13
chore(server): add ssh.dispose to queues

Akirtovskis:
closes  : https://github.com/ledokku/ledokku/issues/300


pradel (approved):
Perfect, we also have to do the same for queries / mutations :)

Akirtovskis:
@pradel added to all mutations and queries as well

pradel (approved):
🚀

PR #308, opened 2021-03-20
chore(server): fix additional feedbacks

Akirtovskis:
Minor fixes that were left from previous git-deployment PR.


server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -34,37 +40,38 @@ export const createAppGithub: MutationResolvers['createAppGithub'] = async (
...
+  });
+
+  const branch = await octokit.repos.getBranch({
+    owner: repoData.owner,
+    repo: repoData.repoName,
+    branch: input.branchName ? input.branchName : 'main',
+  });
+
+  if (!branch.url) {
+    throw new Error("There's no such branch for this repository");
```
pradel, line 57: let|s add the branch name to the error message for easier debugging

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -34,37 +40,38 @@ export const createAppGithub: MutationResolvers['createAppGithub'] = async (
...
   const repoData = getRepoData(input.gitRepoUrl);
-  try {
-    repo = await octokit.repos.get({
-      owner: repoData.owner,
-      repo: repoData.repoName,
-    });
-  } catch (error) {
-    console.log(error);
+
+  const repo = await octokit.repos.get({
```
pradel, line 45: if repo does not exist what will happen here?

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -34,37 +40,38 @@ export const createAppGithub: MutationResolvers['createAppGithub'] = async (
...
+  const randomToken = generateRandomToken(40);
 
   const app = await prisma.app.create({
     data: {
       name: input.name,
       type: 'GITHUB',
       AppMetaGithub: {
         create: {
-          repoUrl: input.gitRepoUrl,
+          repoUrl: `${repoData.owner}/${repoData.repoName}`,
```
pradel, line 72: do we want to change the name of this field?
If we want to apply the github naming we can have 2 fields, one "owner" and one "name"

server/src/graphql/mutations/createAppGithub.ts
Akirtovskis: @pradel after latest changes it will throw error.

server/src/graphql/mutations/createAppGithub.ts
Akirtovskis: @pradel done.

server/src/graphql/mutations/createAppGithub.ts
Akirtovskis: Yeah, I guess it makes sense to have two fields. Will create a migration in couple of min.

pradel (approved):

PR #320, opened 2021-03-24
fix(docs): update command for destroying redis instance

holamendi:
from `dokku postgres:destroy redis` to `dokku redis:destroy ledokku`


pradel:
Thanks @holamendi!

PR #342, opened 2021-04-06
chore(client): migrate terminal component to chakra

pradel:


client/src/ui/components/Terminal.tsx
```diff
@@ -1,17 +1,17 @@
...
+    p="4"
+    boxShadow="lg"
+    borderRadius="lg"
+    color="gray.100"
+    backgroundColor="gray.900"
+    fontSize="sm"
+    fontFamily="mono"
+    lineHeight="1.5"
+    style={{ WebkitFontSmoothing: 'auto' }}
+    {...props}
```
Akirtovskis, line 15: @pradel shouldn't this be `<Box... > {children} </Box>` or will the current approach work as well?

pradel: Actually doing `<Box>{children} </Box>` is the same as doing `<Box children={children} />` so the above is working perfectly as children is present in props :)

Akirtovskis (approved):

PR #357, opened 2021-04-11
feat(client,server): fetching repos/branches from user

Akirtovskis:


server/src/graphql/queries/repositories.ts
```diff
@@ -0,0 +1,131 @@
...
+    iat: Math.round(new Date().getTime() / 1000),
+    exp: Math.round(new Date().getTime() / 1000 + 60),
+    iss:
+      process.env.NODE_ENV === 'production'
+        ? config.githubAppId
+        : process.env.GITHUB_APP__ID,
+  };
+
+  //TODO FIGURE OUT A WAY TO PARSE THE PEM PROPERLY
+  const token = jsonwebtoken.sign(
```
Akirtovskis, line 32: @pradel Note for this to-do do you have a clear idea what should be done to make this work w/o passing hardcoded key?

server/src/graphql/queries/repositories.ts
```diff
@@ -0,0 +1,100 @@
...
+    payload,
+    `-----BEGIN RSA PRIVATE KEY-----
+---- KEY
+-----END RSA PRIVATE KEY-----`,
+    {
+      algorithm: 'RS256',
+    }
+  );
+  const authAsInstallation = await fetch(
```
pradel, line 40: can we use octokit?

server/src/graphql/queries/repositories.ts
```diff
@@ -0,0 +1,100 @@
...
+      headers: {
+        Accept: 'application/vnd.github.machine-man-preview+json',
+        Authorization: `Bearer ${token}`,
+      },
+      method: 'POST',
+    }
+  );
+  const data = await authAsInstallation.json();
+  console.log(data);
+  const responseRepos = await fetch(
```
pradel, line 52: octokit

server/src/graphql/queries/repositories.ts
```diff
@@ -0,0 +1,100 @@
...
+    {
+      headers: {
+        Accept: 'application/vnd.github.machine-man-preview+json',
+        Authorization: `Bearer ${token}`,
+      },
+      method: 'POST',
+    }
+  );
+  const data = await authAsInstallation.json();
+  console.log(data);
```
pradel, line 51: lets remove?

server/src/graphql/queries/repositories.ts
```diff
@@ -0,0 +1,100 @@
...
+      },
+    }
+  );
+
+  const fetchedRepos = await responseRepos.json();
+
+  let repositories = [];
+  let branches = [];
+
+  const map = fetchedRepos.repositories.map(async (r) => {
```
pradel, line 67: so here you are fetching all the branches for all the repos, if 100 repos are returned it will result in 100 HTTP requests made. I think we should have 2 queries: this one which is good and exposes another one that will accept a github repo id and return the branches for this repo.
On the frontend every time the user switch the repo using the select input we would call the branches query and load the values for only one repo. That way this query will resolve much faster :)

server/src/graphql/queries/repositories.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const repos = await octo.request('GET /installation/repositories');
+
+  const repositories = [];
+
+  repos.data.repositories.map((r) => {
+    const repoToPush = {
+      id: r.id,
+      name: r.name,
+      full_name: r.full_name,
```
pradel, line 41: Based on your schema:
```suggestion
      fullName: r.full_name,
```

pradel (approved):

server/src/graphql/queries/branches.ts
```diff
@@ -0,0 +1,53 @@
...
+    auth: installationAuthentication.token,
+  });
+
+  const fetchedBranches = await octo.request(
+    `GET /repos/${user.username}/${repositoryName}/branches`
+  );
+
+  const branches = [];
+
+  fetchedBranches.data.map((b) => {
```
pradel, line 44: as you are using the map function what do you think about this?
```suggestion
  const branches = fetchedBranches.data.map((b) => {
```
and then return the new object inside. Otherwise better use forEach if you prefer this way

server/src/graphql/queries/repositories.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const octo = new Octokit({
+    auth: installationAuthentication.token,
+  });
+
+  const repos = await octo.request('GET /installation/repositories');
+
+  const repositories = [];
+
+  repos.data.repositories.map((r) => {
```
pradel, line 35: same comment here as you use map better to do
```suggestion
  const repositories = repos.data.repositories.map((r) => {
```

PR #358, opened 2021-04-11
feat(server, client): modify-create-github-app-mutation

Akirtovskis:


server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -32,43 +34,42 @@ export const createAppGithub: MutationResolvers['createAppGithub'] = async (
...
+    clientSecret: config.githubAppClientSecret,
   });
 
-  if (!repo) {
-    throw new Error(
-      `No repository found for ${repoData.owner} with name ${repoData.repoName}`
-    );
-  }
-
-  const branch = await octokit.repos.getBranch({
```
pradel, line 56: @Akirtovskis I think we should still verify that the branch does exist

pradel: We should also check that the repoId is valid

server/src/index.ts
```diff
@@ -224,7 +225,9 @@ const typeDefs = gql`
   input CreateAppGithubInput {
     name: String!
     gitRepoUrl: String!
```
pradel, line 227: as we are passing the id we can probably remove this no?

server/src/graphql/mutations/createAppGithub.ts
```diff
@@ -32,43 +34,42 @@ export const createAppGithub: MutationResolvers['createAppGithub'] = async (
...
-      `There's no ${input.branchName} branch or main branch for this repository`
-    );
-  }
+  const repoData = getRepoData(input.gitRepoUrl);
 
   const ssh = await sshConnect();
 
   const dokkuApp = await dokku.apps.create(ssh, input.name);
 
+  const dokkuAuth = await dokku.git.auth({
```
pradel, line 67: do we have to do this here? My understanding is that a token is valid for 24h only, so we should rather call this every time we receive a webhook push event.

server/src/graphql/mutations/createAppGithub.ts
Akirtovskis: @pradel reason I removed these checks is that on the client side we are fetching repo and branch data from the Github app instead of entering manually so the validation becomes redundant.

Do we still need to validate?

pradel: The rule is to never trust the client as I could manually call this endpoint with some fake data.

Akirtovskis: Ok, will add this back then.

server/src/graphql/mutations/createAppGithub.ts
Akirtovskis: Good point, will add this to webhook handler too. I think we should leave it here as well as we need to be authenticated before the first deploy as well.

pradel: Yeah actually we could probably move it in the queue, that way we would have this logic in only one place.

server/src/index.ts
Akirtovskis: @pradel I think passing full name instead of `gitRepoUrl` makes more sense as that's what we use later on the server side. 

Full name is in format  `userName/repoName`

server/src/graphql/mutations/createAppGithub.ts
Akirtovskis: Yeah. moving this to queue makes a lot of sense. 

On it.

pradel (approved):

PR #360, opened 2021-04-17
feat(server): webhooks with new setup

Akirtovskis:


server/src/index.ts
```diff
@@ -473,21 +471,11 @@ app.get('*', (_, res) => {
   );
 });
 
-const debug = createDebug(`webhooks`);
+app.post('/events', async (req, res) => {
+  console.log('received request -----------------------------', req.body);
```
pradel, line 475: Let's change this to debug?

server/src/lib/webhooks/handleWebhooks.ts
```diff
@@ -0,0 +1,47 @@
+import { prisma } from '../../prisma';
+import { Request } from 'express';
+import { createAppAuth } from '@octokit/auth-app';
+import { AppAuthentication } from '@octokit/auth-app/dist-types/types';
+import { config } from '../../config';
+import { deployAppQueue } from '../../queues/deployApp';
+
+export const handleWebhooks = async (req: Request) => {
+  if (!req.body) {
```
pradel, line 9: we also need to verify the request is really coming from github

server/src/lib/webhooks/handleWebhooks.ts
```diff
@@ -0,0 +1,47 @@
...
+  const appsToRedeploy = await prisma.appMetaGithub.findMany({
+    where: {
+      repoId: req.body.repository.id.toString(),
+      repoOwner: req.body.repository.owner.name,
+      githubAppInstallationId: req.body.installation.id.toString(),
+      webhooksSecret: config.githubAppWebhookSecret,
+    },
+  });
+
+  appsToRedeploy.forEach(async (app) => {
```
pradel, line 34: let's do a for of loop instead of forEach as async will be run in parallel with forEach

server/src/lib/webhooks/handleWebhooks.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const installationAuthentication = (await auth({
+    type: 'installation',
+    installationId: req.body.installation.id.toString(),
+  })) as AppAuthentication;
+
+  const appsToRedeploy = await prisma.appMetaGithub.findMany({
+    where: {
+      repoId: req.body.repository.id.toString(),
+      repoOwner: req.body.repository.owner.name,
```
pradel, line 28: do you need the other fields in the findMany? I think just repoId is good enough no?

client/src/modules/webhooks/Webhooks.tsx
```diff
@@ -48,7 +49,7 @@ export const Webhooks = ({ appId }: AppDomainProps) => {
         <GridItem colSpan={8} pt={3}>
           <Text>
             {isSecretVisible
-              ? data.app?.appMetaGithub?.webhooksSecret
+              ? config.githubWebhooksSecret
```
pradel, line 52: @Akirtovskis I think we can remove this totally, the user should not be able too see it and change it.

server/src/index.ts
```diff
@@ -462,7 +460,8 @@ app.get('/runtime-config.js', (_, res) => {
   window['runConfig'] = {
     GITHUB_APP_CLIENT_ID: '${config.githubAppClientId}',
     TELEMETRY_DISABLED: '${config.telemetryDisabled}',
-    GITHUB_APP_NAME: '${config.githubAppName}'
+    GITHUB_APP_NAME: '${config.githubAppName}',
+    GITHUB_APP_WEBHOOKS_SECRET: '${config.githubAppWebhookSecret}'
```
pradel, line 464: with this, you are exposing the secret to all the unauthenticated users, and it's bad idea security-wise 😁

server/src/index.ts
```diff
@@ -472,22 +471,15 @@ app.get('*', (_, res) => {
...
-    debug(`Webhook verification failed for req ${req},`);
-  } else {
-    res.status(200).end();
-    githubPushWebhookHandler(isWebhookVerified);
+app.post('/api/webhooks', async (req, res) => {
+  debug('received request -----------------------------', req.body);
+
+  if (req.header('x-github-event') === 'push') {
+    await handleWebhooks(req);
   }
```
pradel, line 481: shouldn't we return 200 at the end? Also we should use a try catch and return a 500 error in case handleWebhooks return an issue.

pradel (approved):

PR #366, opened 2021-04-24
chore(server): add refreshToken and refreshTokenExpiresIn to user

Akirtovskis:


server/src/graphql/utils.ts
```diff
@@ -47,3 +50,45 @@ export const formatGithubPem = (pem: string) => {
 
   return formattedPem;
 };
+
+export const refreshAuthToken = async (userId: string) => {
+  const user = await prisma.user.findUnique({
+    where: {
```
pradel, line 56: for better perf let's select only the fields we need

server/src/graphql/utils.ts
```diff
@@ -47,3 +50,45 @@ export const formatGithubPem = (pem: string) => {
...
+      id: userId,
+    },
+  });
+
+  let data;
+
+  try {
+    const res = await fetch('https://github.com/login/oauth/access_token', {
```
pradel, line 64: can we use octokit for this?

server/src/graphql/queries/githubInstallationId.ts
```diff
@@ -21,7 +22,21 @@ export const githubInstallationId: QueryResolvers['githubInstallationId'] = asyn
     auth: user.githubAccessToken,
   });
 
-  const installations = await octokit.request('GET /user/installations');
+  let installations;
+
+  try {
```
pradel, line 27: while this is working for this call, it means that we will have to do a try catch in every call that is using the user token. We should abstract the request logic into a function that we can reuse easily )

server/src/graphql/mutations/loginWithGithub.ts
```diff
@@ -77,7 +79,8 @@ export const loginWithGithub: MutationResolvers['loginWithGithub'] = async (
         avatarUrl: githubUser.avatar_url,
         email: email.email,
         githubAccessToken: data.access_token,
-        // TODO save refresh token + refresh_token_expires_in?
+        refreshToken: data.refresh_token,
+        refreshTokenExpiresIn: data.refresh_token_expires_in.toString(),
```
pradel, line 83: Is `refresh_token_expires_in` a date or just an Int? Because if we store it like this we don't know exactly when it will expire. The best would be to change the field to `refreshTokenExpiresAt` which would be a date when the token will expire.

Akirtovskis: @pradel it's an Int. So better to store as an Int?

pradel: Well the int is relative to the time when you store it. Eg: 10 because it will expire in 10 hours.
If you just store 10 there is no way to know when it will expire as the object could have been updated 5 hours or 1 month ago.
Storing the date when it will expire is the best as it can be understood at any time we do a user query in the future.

server/src/graphql/mutations/loginWithGithub.ts
```diff
@@ -71,13 +73,21 @@ export const loginWithGithub: MutationResolvers['loginWithGithub'] = async (
     // Github return an array of emails, we just need the primary email
     const email = emails.find((email) => email.primary);
 
+    // We add token expires in to current date
+    const rn = new Date();
+    const time = rn.getTime();
+    const refreshTokenExpiresIn = new Date(
+      time + data.refresh_token_expires_in
```
pradel, line 80: is the value of refreshTokenExpiresIn minutes or hours?

server/src/graphql/utils.ts
```diff
@@ -47,3 +50,80 @@ export const formatGithubPem = (pem: string) => {
...
+
+    return data;
+  } catch (error) {
+    console.error(error);
+    throw new Error('Request failed');
+  }
+};
+
+export const octoRequestWithUserToken = async (
+  requestData: string,
```
pradel, line 104: can we get better typings? octokit.request is strongly typed, would be great if we could keep it somehow

server/src/graphql/utils.ts
```diff
@@ -47,3 +50,80 @@ export const formatGithubPem = (pem: string) => {
...
+      refreshTokenExpiresIn: true,
+      githubAccessToken: true,
+    },
+  });
+
+  let data;
+
+  const octo = new Octokit({});
+
+  try {
```
pradel, line 70: I am not sure we really need this try catch

server/src/graphql/utils.ts
```diff
@@ -47,3 +50,80 @@ export const formatGithubPem = (pem: string) => {
...
+      },
+      body: JSON.stringify({
+        refresh_token: user.refreshToken,
+        grant_type: 'refresh_token',
+        client_id: config.githubAppClientId,
+        client_secret: config.githubAppClientSecret,
+      }),
+    });
+
+    data = res;
```
pradel, line 84: is this really needed as you return in the try?

server/prisma/schema.prisma
```diff
@@ -8,17 +8,19 @@ generator client {
...
+  id                    String     @id @default(uuid())
+  createdAt             DateTime   @default(now())
+  updatedAt             DateTime   @default(now())
+  username              String     @unique
+  avatarUrl             String
+  email                 String
+  githubId              String     @unique
+  githubAccessToken     String
+  refreshToken          String
+  refreshTokenExpiresIn DateTime
```
pradel, line 20: what about refreshTokenExpiresAt  for the name?

server/src/graphql/utils.ts
```diff
@@ -47,3 +51,74 @@ export const formatGithubPem = (pem: string) => {
...
 };
+
+export const refreshAuthToken = async (userId: string) => {
+  const user = await prisma.user.findUnique({
+    where: {
+      id: userId,
+    },
+    select: {
+      refreshToken: true,
+      refreshTokenExpiresAt: true,
```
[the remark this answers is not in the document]
Akirtovskis: Nop, will remove

PR #392, opened 2021-07-25
feat: dokku installation

akhil-naidu:
### Newly Added
- Colors via `tput` (An inbuilt Linux colour)
- using `Whiptail` ( An inbuilt Linux utility for better TUI experience)
- Gathering sys info, for providing detailed error messages
- Making the script run as root or sudo. (Some providers does not allow passwordless sudo)
- Updating System (The one-click dokku installation in Digital Ocean is so outdated)
- Installing Dokku using the script (If Dokku does not exist)

### Updated
- Intro Comments
- Added Color to echo commands during installation
- Added newly defined functions to the main function

### DIFF
--- a/ledokku-bootstrap.sh
+++ b/ledokku-bootstrap.sh
@@ -86,13 +86,45 @@ function systemUpdate {
   fi
 }
 
-# Check that dokku is installed on the server
-ensure-dokku() {
-  if ! command -v dokku &> /dev/null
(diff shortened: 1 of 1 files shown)


akhil-naidu:
What is wrong with this bot? It is case sensitive!! How can someone start the title with a `non-capital letter`? -_-

PR #393, opened 2021-07-25
feat: improve ledokku installation script

akhil-naidu:
### Newly Added
- Colors via `tput` (An inbuilt Linux colour)
- using `Whiptail` ( An inbuilt Linux utility for better TUI experience)
- Gathering sys info, for providing detailed error messages
- Making the script run as root or sudo. (Some providers does not allow passwordless sudo)
- Updating System (The one-click dokku installation in Digital Ocean is so outdated)
- Installing Dokku using the script (If Dokku does not exist)

### Updated
- Intro Comments
- Added Color to echo commands during installation
- Added newly defined functions to the main function

### DIFF
--- a/ledokku-bootstrap.sh
+++ b/ledokku-bootstrap.sh
@@ -32,7 +32,7 @@ function system-info() {
   DOKKU_SSH_HOST=$(curl -4 ifconfig.co)
   OS=$( $(compgen -G "/etc/*release" > /dev/null) && cat /etc/*release | grep ^NAME | tr -d 'NAME="' || echo "${OSTYPE//[0-9.]/}")
 
-  ## Dokku Configuration Varibales
+  ## Dokku Configuration Variables
   LATEST_DOKKU_VERSION="0.24.10"
   MINIMUM_DOKKU_VERSION="0.24.0"
 }
--- a/ledokku-bootstrap.sh
+++ b/ledokku-bootstrap.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# This script is used to configure the essentials of t2d.
+# This script is used to configure the essentials of ledokku.
 ## 1 => Check whether the program/application, "Whiptails" exists or not.
 ## 2 => Making sure that the script is runing with root permissions.
 ## 3 => Update and Upgrade the VPS.
@@ -27,9 +27,9 @@ function define-colors {
 
 function system-info() {
   
-  # Finding Information about your device
+  # Finding Information about the server
   ## Basic VPS info
-  ## Should exist in every script
+
   DOKKU_SSH_HOST=$(curl -4 ifconfig.co)
   OS=$( $(compgen -G "/etc/*release" > /dev/null) && cat /etc/*release | grep ^NAME | tr -d 'NAME="' || echo "${OSTYPE//[0-9.]/}")
 }
@@ -51,7 +51,7 @@ function check-whiptail() {
 
 function check-root() {
   
-  #Check root and if not root take permissions (Some providers doesnot support passwordless sudo)
+  # Check root and if not root take permissions (Some providers doesnot support passwordless sudo)
   ## It is always better to do this.
   ## We will not face any further issues, during any sort of compulsory sudo commands; like the case for installing plugins in Dokku or Giving permissions to our scripts
   if [ "$(whoami)" == "root" ] ; then
@@ -97,8 +97,8 @@ function ensure-dokku() {
   if which dokku >/dev/null; then
       echo "${GREEN}Dokku Exists${END}"
       # Checking Dokku vesrion and comparing it with the latest Version
-      # In case of version changes in dokku, we need to change this varibale.
-      # We can also change the LATEST_DOKKU_VERSION to PREFFERED_DOKKU_VERSION
+      # In case of version changes in dokku, we need to change this varibale: LATEST_DOKKU_VERSION.
+      # We can also rename the variable => LATEST_DOKKU_VERSION to PREFFERED_DOKKU_VERSION
       
       EXISTING_DOKKU_VERSION="$(dokku version)"
       LATEST_DOKKU_VERSION="dokku version 0.24.10"
@@ -146,7 +146,7 @@ function ensure-dokku() {
 }
 
 # Check if dokku redis plugin is intalled and otherwise install it
-install-redis() {
+function install-redis() {
   if sudo dokku plugin:installed redis; then
     echo "=> ${GREEN}Redis plugin already installed skipping${END}"
   else
@@ -156,7 +156,7 @@ install-redis() {
 }
 
 # Check if dokku postgres plugin is intalled and otherwise install it
-install-postgres() {
+function install-postgres() {
   if sudo dokku plugin:installed postgres; then
     echo "=> ${GREEN}Postgres plugin already installed skipping${END}"
   else
@@ -165,7 +165,7 @@ install-postgres() {
   fi
 }
 
-main() {
+function main() {
   define-colors
   system-info
   check-whiptail
--- a/ledokku-bootstrap.sh
+++ b/ledokku-bootstrap.sh
@@ -2,13 +2,13 @@
 
 # This script is used to configure the essentials of ledokku.
 ## 1 => Check whether the program/application, "Whiptails" exists or not.
-## 2 => Making sure that the script is runing with root permissions.
+## 2 => Making sure that the script is running with root permissions.
 ## 3 => Update and Upgrade the VPS.
 ## 4 => Check whether the program/application, "Dokku" was Installed or not.
 ## 5 => Upgrading dokku to the latest version.
 ## 6 => Check whether the Plugins are installed or not (Redis, Postgres)
-## 7 => Install Redis and Postgress, if they were not installed
-## 8 => Install Ledokku
+## 7 => Install Redis and Postgres, if they were not installed
+## 8 => Install ledokku
 ## ? => Update Script (Will be added later)
 
 set -e
@@ -54,7 +54,7 @@ function check-whiptail() {
 
 function check-root() {
   
-  # Check root and if not root take permissions (Some providers doesnot support passwordless sudo)
+  # Check root and if not root take permissions (Some providers does not support password less sudo)
   ## It is always better to do this.
   ## We will not face any further issues, during any sort of compulsory sudo commands; like the case for installing plugins in Dokku or Giving permissions to our scripts
   if [ "$(whoami)" == "root" ] ; then
@@ -96,15 +96,15 @@ function system-update {
 
 function ensure-dokku() {
   
-  # Confirming the existance of Dokku
+  # Confirming the existence of Dokku
   ## If exists => check if existing and latest dokku versions are same => If same version => Continue to next step
-  ###  In case of different dokku versions => Promot a (warning & dokku upgrade) dialog box (Now it is upto the user to update or skip)
+  ###  In case of different dokku versions => Prompt a (warning & dokku upgrade) dialog box (Now it is up to the user to update or skip)
   ## If dokku does not exits, take permission to download Dokku.
   if which dokku >/dev/null; then
       echo "${GREEN}Dokku Exists${END}"
       # Checking Dokku version and comparing it with the latest Version
-      # In case of version changes in dokku, we need to change this varibale: LATEST_DOKKU_VERSION.
-      # We can also rename the variable => LATEST_DOKKU_VERSION to PREFFERED_DOKKU_VERSION
+      # In case of version changes in dokku, we need to change this variable: LATEST_DOKKU_VERSION.
+      # We can also rename the variable => LATEST_DOKKU_VERSION to PREFERRED_DOKKU_VERSION
       
       EXISTING_DOKKU_VERSION="$(dokku version | awk '{print $3}')"
 
@@ -116,9 +116,9 @@ function ensure-dokku() {
           # Continue the script
           echo "${GREEN}Awesome! You have the latest dokku version${END}"
         else 
-          # Promot upgrade warning (upgrade or skip)
+          # Prompt upgrade warning (upgrade or skip)
           whiptail --title "Warning !!" --msgbox "Read carefully before proceeding:\n\nYou are currently using dokku version: ${EXISTING_DOKKU_VERSION} but the latest dokku version was: ${LATEST_DOKKU_VERSION}\n\nIn the next dialog box, you can upgrade your dokku or skip to ledokku installation \n\nFor more info check the dokku CHANGELOG before doing the upgrade: https://github.com/dokku/dokku/releases" 20 60
-          # Promt for upgrade
+          # Prompt for upgrade
           if (whiptail --title "Upgrading Dokku" --yes-button "Upgrade" --no-button "Skip"  --yesno "Would you like to upgrade your Dokku?" 10 60) then
               echo "${YELLOW}You chose Upgrade.${END}"
               # Upgrade Dokku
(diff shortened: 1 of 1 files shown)


ledokku-bootstrap.sh
```diff
@@ -1,39 +1,162 @@
...
+  BLUE="$(tput setaf 123)" # ${BLUE}
+  END="$(tput setaf 7)" # ${END
+}
+
+function system-info() {
+  
+  # Finding Information about your device
+  ## Basic VPS info
+  ## Should exist in every script
+  IP="$(ifconfig | grep broadcast | awk '{print $2}')"
```
pradel, line 33: is this used? We already get the ip line 166

pradel (commented):
Thanks for the pr, a lot of good ideas inside. As general feedback, dokku supports `Ubuntu 18.04/20.04 x64, Debian 9+ x64 or CentOS 7 x64` so are you sure all the changes made on this pull request will work when installing ledokku on such systems?

akhil-naidu:
Here is a [video that demonstrates the usage of this script](https://www.youtube.com/watch?v=Ajbek9xxwHc). With this new script, there is no need for the users to have a pre-installed dokku server.

It checks whether dokku was installed or not. If it was not able to find an existing dokku, it will install dokku by promoting a user verification for proceeding.  Also, if there is an outdated dokku version, it will prompt permission to update; the user can choose to update or skip.

ledokku-bootstrap.sh
akhil-naidu: If there are more than 2 IP addresses, it can be useful, or else we can remove this.

ledokku-bootstrap.sh
```diff
@@ -1,39 +1,162 @@
...
+
+function ensure-dokku() {
+  
+  # Confirming the existance of Dokku
+  ## If exists, promt for update
+  ## If not exits Let him download.
+  if which dokku >/dev/null; then
+      echo "${GREEN}Dokku Exists${END}"
+      # Promt for update
+      if (whiptail --title "Updating Dokku" --yes-button "Update" --no-button "Skip"  --yesno "Would you like to update your Dokku?" 10 60) then
```
pradel, line 100: I am not sure about prompting users to update dokku here. How to upgrade dokku is changing from one version to another so we do not cover all the versions with this approach. It's better that the user takes care of this manually and reads the dokku Changelog notes to see what might break during the upgrade.

ledokku-bootstrap.sh
```diff
@@ -1,39 +1,162 @@
...
+          echo "${YELLOW}Now you have the latest Version of Dokku${END}"
+          # Dokku Updated
+      else
+          echo "${Yellow}You chose to skip dokku updates.${END}"
+          # Dokku Update skipped
+      fi
+  else
+      echo "${RED}Dokku does not exist${END}"
+      # Show messagebox and make it mandatory to download and install dokku
+      whiptail --title "Unable to Detect Dokku" --msgbox "If you want to insatll your app using t2d, it is madatory to install Dokku. So, I would like to install Dokku on behalf of you." 10 60
```
pradel, line 121: What is t2d?
Also typo on `insatll`
Offering to install dokku is a great idea!

akhil-naidu:
> Thanks for the pr, a lot of good ideas inside. As general feedback, dokku supports `Ubuntu 18.04/20.04 x64, Debian 9+ x64 or CentOS 7 x64` so are you sure all the changes made on this pull request will work when installing ledokku on such systems?

Yes, the scripts redirect to the official dokku installation; If there are any dependency errors, the dokku script will tackle them(like OS and others) and even if it persists, it will provide the necessary error.

ledokku-bootstrap.sh
pradel: We were using ifconfig but sometimes it reported the internal ip and using the approach line 166 was proven more stable across all installs so it's better to keep it that way :)

ledokku-bootstrap.sh
akhil-naidu: Yes, we can do that, but according to the present script, it will update to the latest version.

akhil-naidu: While using ledokku, if we can force users to use the latest version, then we will not have any conflicts. 

> Keeping that in mind, I'm actually promoting such update requests.

ledokku-bootstrap.sh
akhil-naidu: **Edit 01**: So, I will comment out this, in the next commit before we can merge.

anything else?

ledokku-bootstrap.sh
akhil-naidu: **Edit 02**: I will clear these typos.

t2d => Ledokku
insatll => Install

ledokku-bootstrap.sh
akhil-naidu: ### What should we do about this? 
- Keep it to the latest and stable version for now and we added support for individual versions in the later stage.
 ** or**
- Uncomment those and make it manual(for the user to decide)

---
### My Opinion
If we can force all the ledokku users to one version, it would be easy for us to maintain, rather than leaving it to the discretion of the users.

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,174 @@
...
+      fi
+  else
+      echo "${RED}Dokku does not exist${END}"
+      # Show messagebox and make it mandatory to download and install dokku
+      whiptail --title "Unable to Detect Dokku" --msgbox "If you want to insatll your app using ledokku, it is madatory to install Dokku. So, I would like to install Dokku on behalf of you." 10 60
+      wait
+      echo "${YELLOW}Downloading Dokku from its Official Repository${END}"
+      wget https://raw.githubusercontent.com/dokku/dokku/v0.24.10/bootstrap.sh
+      wait
+      sudo DOKKU_TAG=v0.24.10 bash bootstrap.sh &
```
akhil-naidu, line 127: We can actually make this `DOKKU_TAG` to have defaulted at least version and also give the users an option to change it to a specified version.

This specified version can be an `input from user` or `choose among the supported version, we can offer`

akhil-naidu (commented):
What do you say?

ledokku-bootstrap.sh
pradel: > If we can force all the ledokku users to one version, it would be easy for us to maintain, rather than leaving it to the discretion of the users.

Agree with this, just to make it more safe I would suggest the following changes:
- only print the upgrade message if user is not already using the latest version
- add a message saying something like this:
```
You are currently using Dokku vXXX, would you like to update Dokku to vXXX?
(in red) Please make sure you checked the dokku CHANGELOG before doing the upgrade https://github.com/dokku/dokku/releases
```

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,174 @@
...
+  # Checking if whiptail is available or not
+  if which whiptail >/dev/null; then
+      echo "${GREEN}whiptail exists${END}"
+      # Continue the script
+  else
+      echo "${RED}whiptail does not exist${END}"
+      echo "Install whiptail and re-run the script, your OS is ${OS}"
       exit
+      # As I already know the OS, I can also automate this process, but will save it for later.
+      # This is just a matter of finiding all the possible OS people might use and writing a if statement.
```
pradel, line 49: Typo `finiding `

akhil-naidu: let's figure out all the typos and edit at the end rather than making simultaneous commits for individual typos.

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,174 @@
...
+function ensure-dokku() {
+  
+  # Confirming the existance of Dokku
+  ## If exists, promt for update
+  ## If not exits Let him download.
+  if which dokku >/dev/null; then
+      echo "${GREEN}Dokku Exists${END}"
+      # Promt for update
+      if (whiptail --title "Updating Dokku" --yes-button "Update" --no-button "Skip"  --yesno "Would you like to update your Dokku?" 10 60) then
+          echo "${YELLOW}You chose Update.${END}"
```
pradel, line 102: Let's remove this one as it seems redundant

ledokku-bootstrap.sh
akhil-naidu: #### Typo, Edit 01
finiding => finding
a if   => an if

ledokku-bootstrap.sh
akhil-naidu: Sure!

ledokku-bootstrap.sh
pradel: To limit the number of interactions a user has to do let's do it the same way as we do for `LEDOKKU_TAG`, if the user sets it before the installation script is run we use the one provided by the user, otherwise we default to v2.24.10

akhil-naidu: As we removed the upgrade option, let's keep it to the latest version and remove user interaction. 

Rather than creating this DOKKU_TAG as script variables, let's make DOKKU_TAG a global variable in the entire repository. This way, if the user will have the ability to change it, in case if he prefers.

pradel: Just saw you removed the upgrade option, but what do you think about this instead? https://github.com/ledokku/ledokku/pull/393#discussion_r676130185

ledokku-bootstrap.sh
akhil-naidu: So, you would like me to add a warning message (about the outdated version) and should no prompt the option to update within the script.

Am I catching up with you?

pradel: Warning message so user understand that there is a risk by updating, and then ask him if he wants to update so keep the prompt

akhil-naidu: So now, this is what I'm going to add now.

check dokku => if installed, check the version => if matched with the version we desire => we will pass it to the next command. 

If the installed versions are not matching, I will display the version used by the user and the latest dokku version with a warning message. And also provide the user with the ability to update to the latest dokku version and to also to skip this step.

pradel: SOunds perfect 👌

akhil-naidu: Check it and let me know if we need any changes.

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,187 @@
...
+            # Dokku Updated
+        else
+            echo "${YELLOW}You chose to skip dokku updates.${END}"
+            # Dokku Update skipped
+        fi
+      fi
+  else
+      echo "${RED}Dokku does not exist${END}"
+      # Show messagebox and make it mandatory to download and install dokku
+      whiptail --title "Unable to Detect Dokku" --msgbox "If you want to insatll your app using ledokku, it is madatory to install Dokku. So, I would like to install Dokku on behalf of you." 10 60
```
Akirtovskis, line 135: @akhil-naidu typo in the first `install`

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,187 @@
...
       exit
   fi
 }
 
+function system-update {
+
+  # Updating and Upgrading system
+  ## If dokku was installed, it will be automatically updated to the latest version
+  ## Staying up to date is always good
+  if (whiptail --title "Update and Upgrade System " --yes-button "Yes" --no-button "Skip"  --yesno "Do you wish to Update packges and Upgrade your system?" 10 60) then
```
Akirtovskis, line 70: @akhil-naidu 

Firstly, *packages typo. 

And secondly, could you please make the message more verbose, so it is clear that we are talking about the packages related to dokku/ledokku. Otherwise first trying this out I was not sure what kind of upgrades are we talking about here.

ledokku-bootstrap.sh
akhil-naidu: Edit 01: I will take care of that typo.

let me also know about other issues.

ledokku-bootstrap.sh
akhil-naidu: It is updating APT(Advanced Package Tool) => nothing but `sudo apt update`

Edit 02: I will update the typo and will be more specific.

akhil-naidu: Let me know if there are any issues in the flow of the logic. Meanwhile, I will check for typos and other instructions.

Akirtovskis: @akhil-naidu Apart from typos everything else looks good to me and just worked on a fresh Digital Ocean server.

ledokku-bootstrap.sh
```diff
@@ -69,15 +208,15 @@ main() {
   dokku config:set ledokku JWT_SECRET="${JWT_SECRET}"
 
   # Now it's finally time to install ledokku
-  echo "=> Installation"
+  echo "=> ${YELLOW}Installing ledooku${END}"
   dokku git:from-image ledokku ledokku/ledokku:${LEDOKKU_TAG}
```
akhil-naidu, line 212: Dokku has introduced the `git:from-image <app> <docker-image>` from **dokku v0.24.0** 

The digital ocean marketplace provide a **dokku v0.21.4** => we should force users to have a minimum of dokku v0.24.0

akhil-naidu (commented):
Should fix this version issue.

akhil-naidu:
### About the previous commit.
The above `error fixed` commit consists of the following list of commits.
- Globally defining **LATEST_DOKKU_VERSION** and **MINIMUM_DOKKU_VERSION**
   - This helps us update the latest dokku version with ease.
   - This will also help us in bumping the minimum required version.
   - Changing these variables will change the logic of the dokku installation. (Accordingly)
- It consists of dokku installation logic based upon `3 conditions`
   - If dokku was not installed in the system, we will install the **latest version of dokku**
   - If there is a pre-installed dokku with compatible versions(in b/w ***MINIMUM_DOKKU_VERSION*** and ***LATEST_DOKKU_VERSION***)
      > The user will also have the ability to upgrade his dokku to the latest version or skip dokku upgrade. (With a warning message)
   - If the pre-installed dokku is a ledokku incompatible version( version **less than *MINIMUM_DOKKU_VERSION***),
      > The user will have the ability to upgrade to the latest dokku version or end the ledokku script. (With a warning message)

ledokku-bootstrap.sh
akhil-naidu: fixed: https://github.com/ledokku/ledokku/pull/393/commits/ac2b0e3906a578a7bcc991683e4562d5c45deb98

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,223 @@
...
+}
+
+function system-info() {
+  
+  # Finding Information about the server/VPS
+  ## Basic VPS info
+  DOKKU_SSH_HOST=$(curl -4 ifconfig.co)
+  OS=$( $(compgen -G "/etc/*release" > /dev/null) && cat /etc/*release | grep ^NAME | tr -d 'NAME="' || echo "${OSTYPE//[0-9.]/}")
+
+  ## Dokku Configuration Varibales
```
pradel, line 35:
```suggestion
  ## Dokku Configuration Variables
```

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,223 @@
...
+
+function ensure-dokku() {
+  
+  # Confirming the existance of Dokku
+  ## If exists => check if existing and latest dokku versions are same => If same version => Continue to next step
+  ###  In case of different dokku versions => Promot a (warning & dokku upgrade) dialog box (Now it is upto the user to update or skip)
+  ## If dokku does not exits, take permission to download Dokku.
+  if which dokku >/dev/null; then
+      echo "${GREEN}Dokku Exists${END}"
+      # Checking Dokku vesrion and comparing it with the latest Version
```
pradel, line 105:
```suggestion
      # Checking Dokku version and comparing it with the latest Version
```

ledokku-bootstrap.sh
```diff
@@ -1,52 +1,223 @@
...
+              echo "${YELLOW}Upgraded to ${GREEN} $LATEST_DOKKU_VERSION ${END}"
+              # Dokku Updated
+          else
+              echo "${YELLOW}You chose to skip dokku upgrades.${END}"
+              # Dokku Update skipped
+          fi
+        fi
+      else
+        echo "Your dokku version is ${EXISTING_DOKKU_VERSION} => and for ledokku compatibility dokku version should be between ${MINIMUM_DOKKU_VERSION} and ${LATEST_DOKKU_VERSION}"
+        # Promot upgrade to latest dokku version (upgrade or exit)
```
pradel, line 143:
```suggestion
        # Prompt upgrade to latest dokku version (upgrade or exit)
```

pradel (approved):

Akirtovskis (approved):

Akirtovskis:
@akhil-naidu good job! 🙏🏻 🚀 🏆

PR #412, opened 2022-03-27
docs: update uninstall.md

troya2:
Additional work is needed to completely remove ledokku (and allow reinstalling it, if desired).  Without removing those additional files and directories, the install script will fail if reinstalling.


pradel (approved):
Thanks for the pr!

